PR #2, opened 2015-05-26
Control MCA parameters through environement vars

pablooliveira:
The environment variable `VERIFICARLO_MCAMODE` controls the arithmetic error
mode. It accepts the following values:

 * `MCA`: (default mode) Montecarlo Arithmetic with inbound and outbound errors
 * `IEEE`: the program uses standard IEEE arithmetic, no errors are introduced
 * `PB`: Precision Bounding inbound errors only
 * `RR`: Random Rounding outbound errors only

The environment variable `VERIFICARLO_PRECISION` controls the virtual
precision used for the floating point operations. It accept an integer value
that represents the number of significant digits. The default value is 23.

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,6 +80,22 @@ If you only wish to instrument a specific function in your program, use the
    $ verificarlo *.c -o ./program --function=specificfunction
 ```
 
+## MCA Configuration Parameters
+
+Two environement variables control the Montecarlo Arithmetic parameters.
+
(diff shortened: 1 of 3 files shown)


pablooliveira:
@titeup Could you please review this and merge if Travis CI build is successful ? Thanks !

pablooliveira:
@titeup, thanks closing.

PR #3, opened 2015-10-22
Change default MCA precision to 53 bits

pablooliveira:
As discussed it would be nice that the default MCA precision is 53 bits (double precision).


pablooliveira:
@titeup could you please review and merge this ?

titeup:
ok

PR #4, opened 2015-10-26
Refactor verificarlo mca interface to support multiple MCA backends

pablooliveira:
We want to support different MCA backends (to optimize speed of operation). This requires redesigning the mca library hooks in verificarlo.

I open this PR to track the work on this issue.


pablooliveira:
@titeup, this PR is complete and paves the way for other backend such as quadlib. Could you please review and if ok merge it ? Thanks !

PR #5, opened 2015-10-28
New quad version of the mcalib

titeup:
initial commit of the quad implementation, still mpfr operations as default for quad

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -106,6 +106,7 @@ AC_CONFIG_FILES([Makefile
                  src/Makefile
                  src/libvfcinstrument/Makefile
                  src/libmca-mpfr/Makefile
+		 src/libmca-quad/Makefile
                  tests/Makefile])
 
 AC_CONFIG_FILES([verificarlo], [chmod +x verificarlo])
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -1,1 +1,1 @@
-SUBDIRS=libvfcinstrument libmca-mpfr
+SUBDIRS=libvfcinstrument libmca-mpfr libmca-quad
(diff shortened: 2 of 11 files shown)


titeup:
Hi guys, 
Tests looks ok. We still have to check together the build system ;)
Cheers,
Eric

titeup:
Tested the perf on kahan, I observed a speedup of 3 for the new maclib based on quad.
Add some minor improvements o the last version but it only return a few percent SU.
New idea for optimizations are welcome...

pablooliveira:
Very nice work !! Thanks Eric. I have a few very minor nitpicks :
 * please avoid non-informative git log messages such as "It Works !", it makes history hard to follow afterwards :-) You can always use git rebase -i to clean up your local history once a PR is finished.
 * I think the switch(qop) at lines mcalib.c#L194 and mcalib.c#L235 could be factorized
 * the function rexp reads a bit like dark magic to me :-) Some comments on the magic ix constant would help

Since they are really minor tweaks, I'm merging this as it is. Congrats on the x3 speedup.

titeup:
For qop, I prefer to let it as it is since I will probably do another version using double precision for the single precision type. 
For the 'magic' constant, sorry to disappoint you but there is no real magic, even at christmas time ;)  it is just a mask to remove the sign bit, the mantissa disappear with the shift operator just after, then we substract the complement, et voila! I will comment.
I know you already merged, but we might have check the build system first to allow mpfr or quad usage using either a variable at compile time, or at runtime. Anyway, the current master branch seems to work well and faster than the original mpfr version, nobody will really complain I suppose ;)

titeup:
BTW, thanks for merging !

PR #6, opened 2015-11-16
Use double for single operation

titeup:
Hi, 

Forgot to fork before working, so I crete a new branch and create a pull request from there.
I hope it is an 'ok' procedure ;)

Cheers,

Eric

### DIFF
--- a/src/libmca-quad/mcalib.c
+++ b/src/libmca-quad/mcalib.c
@@ -58,6 +58,13 @@ int 	MCALIB_T		    = 53;
 #define QMUL 3
 #define QDIV 4
 
+//possible dop values
+#define DADD 1
+#define DSUB 2
+#define DMUL 3
+#define DDIV 4
+
+
 static float _mca_sbin(float a, float b, int qop);
 
 static double _mca_dbin(double a, double b, int qop);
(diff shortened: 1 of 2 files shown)


titeup:
Hi,

DO NOT MERGE!

Info: I commited the version with the bug, I have no idea yet way it is failling on the function test, facts:
- I only change single prec wrapper and this test is double
- When testing my last changes adding printf to the operator for single prec management, the test works. (this is why Travis is happy while it shouldn't) => DO NOT MERGE!
-This function with printf are not even called => might be a mem alloc problem solved by having allocated memory for the printf format string
Still working on it...

titeup:
Actually I guessed what happen while writing the last comment ;) Bug fixed ;)
However I still do not understand why this specific test failled because of it... And I will probably never know...

titeup:
Last commit tonight.
I finally replaced the quad multiplication by bits operation since one of the operand is a power of 2.
However I suppose because I don't have ieee roundind and conversion during the operation wich means that the lowest bits (up to ~4 empirically)might change in both direction. This must be checked and validated.
Good news however, the final speed up on ingle precision is 35, and 4 on double precision (formerly 3 and 3)

pablooliveira:
I'm sorry ! I messed up when trying to contribute to your PR. I accidentally deleted this branch. I have restored it but apparently github does not allow to reopen a PR once its original branch has been deleted... We can continue discussion at #7.

PR #7, opened 2015-11-17
Quad double double float

pablooliveira:
Reopened the original PR #6 from Eric which I accidentally closed while contributing a patch. Discussion can continue here... Sorry !


titeup:
Hi,
code cleaning (some function as becomed useless after being inlined by end to factorize some operation) and I also forced compiler inlining when relevant with inline keyword.
final SU using test_kahan on double 4.2
on single 35.2

titeup:
Hi Pablo,

I did bug fixes in all versions of libmca about exp computation.
double-simple and mpfr were using frexp and its mpfr equivalent to get the exponent, which is not giving the ieee exp but an exp corresponding to a normalization in the interval [1/2 1[
So I changed it to have the ieee exponent and remove the +1 in the noise exponent computation

titeup:
still something weird. We will have to cotinue the investigation..

titeup:
I updates test_per_function which was failing with the shift in verificarlo_precision effect with the updated backend. 
@pablooliveira  Can you check that to?

pablooliveira:
Test seems fine. There are some small differences between the backend, but I do not think they are statistically significant. I'm merging this, when Travis checks out !

Support of denormalized values in the quad backend still needs work, I'm opening a ticket in that regard.

:tada:

titeup:
Should we change the version number as we discussed?

pablooliveira:
Yep I will bump the version to 0.1.0 and update the documentation with info about the new backend.

PR #9, opened 2015-12-31
Simpe test to check distribution after a MCA sum operation

pablooliveira:
This test is useful to compare the statistical distribution of the different
backends.


titeup:
Looks ok to me, however I still have one question about the test results before merging. (see our emails)

titeup:
ok

PR #10, opened 2016-01-04
Document that QUAD backend uses round-to-zero in MCA ops

pablooliveira:


pablooliveira:
I'm merging this PR which is only documentation.

PR #39, opened 2016-12-11
New version release 0.2.0

pablooliveira:
[v0.2.0] 2016/12/09
      * Support for LLVM up to 3.8
      * Support for standalone installation with make install
      * Experimental VTK post-processing
      * Fix bug with unsupported fortran extensions

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,5 +16,5 @@ install:
   - export LD_LIBRARY_PATH=/usr/local/lib:$LD_LIBRARY_PATH
 
 script:
-  - make check
+  - make installcheck
   - for i in $(find tests/ -maxdepth 1 -type d -name 'test_*' | sort ); do echo "************** TEST $i"; cat $i/test.log; done;
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,5 +1,11 @@
 Verificarlo CHANGELOG
 
+[v0.2.0] 2016/12/09
+    * Support for LLVM up to 3.8
+    * Support for standalone installation with make install
+    * Experimental VTK post-processing
+    * Fix bug with unsupported fortran extensions
+
 [v0.1.0] 2016/01/06
     * QUAD backend performance has been thoroughly optimized
     * QUAD backend now computes MCA operations on float using doubles
--- a/Makefile.am
+++ b/Makefile.am
@@ -1,5 +1,5 @@
 ACLOCAL_AMFLAGS=-I m4
-SUBDIRS=src/ tests/
+SUBDIRS=src/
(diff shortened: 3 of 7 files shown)


titeup:
testing...

titeup:
from a new clone:
make check has no effect weird, I will look closer.
Note when I remove the previous one:
concerning the make install, some local files are created with sudo writes whcih makes necessary the sudo command to clean the directory. Can't we just have only the necessary file written by roo?

titeup:
and a funny message in the terminal:
make[1]: Leaving directory '/home/epetit/verificarlo_master/verificarlo'
echo "Tests should be run after install with make installcheck"
Tests should be run after install with make installcheck

.travis.yml
```diff
@@ -16,5 +16,5 @@ install:
   - export LD_LIBRARY_PATH=/usr/local/lib:$LD_LIBRARY_PATH
 
 script:
-  - make check
```
titeup: Ok, but make check still runs with no error, but no check.
Can we alias?
Should we modify the doc, I do not see this modification in the list.

CHANGELOG
```diff
@@ -1,5 +1,11 @@
 Verificarlo CHANGELOG
 
+[v0.2.0] 2016/12/09
+    * Support for LLVM up to 3.8
+    * Support for standalone installation with make install
```
titeup: For make install in sudo, did you check that clean/distclean and directory removal are still possible for the user, I had some 'root' file in my user directory

Makefile.am
```diff
@@ -12,3 +12,9 @@ verificarlo: verificarlo.in Makefile
 	mv $@.tmp $@
 
 CLEANFILES = verificarlo
+
+check:
+	echo "Tests should be run after install with make installcheck"
```
titeup: the output is:
echo message
message

while we expect just:
message

pablooliveira:
Thanks for the feedback.

1) I replaced "make check" by "make installcheck" because that is the semantics of the two targets in autotools. Make check is used for in source-tree tests. This was a problem, because our tests now require installing the libraries with "make install". If we kept the old target "make check"; when building the debian package a failure was triggered.

Yes, the documentation has been changed to reflect that.

2) About the makedist target. I works well in my end; no "root" owned file created. Are you sure you typed "make" and then "sudo make install". If you type directly "sudo make install" without first typing "make"; the build is performed as root and then generated files are created as root...

3) The "echo" thing is easy to fix if it annoys you. I will add a patch for it...

pablooliveira:
Thanks for the review :party:

PR #42, opened 2017-01-23
Add multiple function instrument

yohanchatelain:


pablooliveira:
Thanks for this @yohanchatelain. As we discussed it would be nice that you do a clean up of your commit.

For example you can use,
$ git rebase -i 2d1f97f

to rework you PR. In particular please remove commit 02c94cc, which makes us lose history.
And you can squash commits 	30c797c, 538d696, 	0806384 and cdbcae9.

Moreover you seem to have some conflicting files.

Thanks :smile:

pablooliveira:
Looking at your commits more closely, I see that 0806384 reverts the changes in 	02c94cc. So there is no need to remove 02c94cc; just squash them together during the rebase.

src/libvfcinstrument/libVFCInstrument.cpp
```diff
@@ -52,8 +60,25 @@ namespace {
...
         static char ID;
         StructType * mca_interface_type;
-
-        VfclibInst() : ModulePass(ID) {}
+        std::set<std::string> SelectedFunctionSet;
+
+        VfclibInst() : ModulePass(ID) {
+	  if (not VfclibInstFunctionFile.empty()) {
+	    std::string line;
+	    std::ifstream loopstream (VfclibInstFunctionFile.c_str());
```
pablooliveira: Add a small comment here explaining what is going on; such as "Parsing command line arguments to known which functions must be instrumented. The chosen functions are collected in SelectedFunctionSet."

PR #45, opened 2017-04-24
Quad backend

titeup:
Hello,

If I made no mistake, this should be finally the final version to merge with master.
Backend has been tested, there is no diffrences when the seeds are fixed with the mpfr backend.
Maybe special values and subnormal would require more testing, but we have to build a test for that.
Can you check if everything is correct Pablo?

Thanks, cheers,

Eric


titeup:
All checks have passed, that's a good news ;)

pablooliveira:
Fantastic !

PR #46, opened 2017-04-24
New backend bitmask and report pass:

yohanchatelain:
* Backend bitmask: apply bitmask to each floating-points operations
                The bitmask depends on the VERIFICARLO_PRECISION
                - Three modes by using VERIFICARLO_BITMASK_MODE environment variable
                        - ZERO: (default) apply 0 to pseudo-mantissa
                        - INV: apply 1 to pseudo-mantissa
			- RAND: apply random bits to pseudo-mantissa

        * Report pass: output a csv file containing information about code by using --report-level=<level>
                - Multiple levels:
                        - module
                        - function
                        - basicblock
                 - An option to print the percentage of type of operation (+,-,*,/) (--print-percentage)


titeup:
Maybe you should add a comment on the explanation of the RAND bitmask backend to say it is using a finite bitfield of ranom bits.
I have seen you add a test for the bitmask backend that was excuted by travis, that's great!
I reviewed the changes but HAVEN'T tested your branch yet, looks ok to me. I'll play around with your branch tonight.

titeup:
ok tested it, works fine,  but ( ;) ) the test is not covering the RAND mode. I'll try to figure something out rapidly with the mpfr/quad test to validate it, but I think it would need more thinking to be done properly in a later release

titeup:
export VERIFICARLO_BACKEND="BITMASK"
        export VERIFICARLO_MODE="RAND"
        ./test > out_bitmask.rand.orig
        cat out_bitmask.rand.orig | grep 'TEST>' | cut -d'>' -f 2 > out_bitmask.rand

and the result:
epetit@majorque:~/verificarlo/tests/test_backends$ cat out_bitmask.rand.orig 
TEST>inf
TEST>inf
TEST>inf
TEST>inf
TEST>inf
TEST>inf
TEST>inf
TEST>inf
TEST>inf
TEST>inf
...

Did I miss something?

src/libbitmask/bitmasklib.c
```diff
@@ -0,0 +1,239 @@
+// The Monte Carlo Arihmetic Library - A tool for automated rounding error
```
pablooliveira: Please change the header, this is not MCALib. Also update copyright to 2017.

src/libbitmask/bitmasklib.c
```diff
@@ -0,0 +1,239 @@
...
+  return dmask;
+}
+
+static void _mca_seed(void) {
+  return;
+}
+
+/******************** MCA ARITHMETIC FUNCTIONS ********************
+* The following set of functions perform the MCA operation. Operands
+* are first converted to MPFR format, inbound and outbound perturbations
```
pablooliveira: you are not using MPFR

src/libbitmask/libbitmask.h
```diff
@@ -0,0 +1,25 @@
+/********************************************************************************
+ *                                                                              *
+ *  This file is part of Verificarlo.                                           *
+ *                                                                              *
+ *  Copyright (c) 2015                                                          *
```
pablooliveira: 2017

src/libvfcinstrument/libVFCInstrument.cpp
```diff
@@ -57,6 +57,13 @@ static cl::opt<bool> VfclibInstVerbose("vfclibinst-verbose",
 				       cl::desc("Activate verbose mode"),
 				       cl::value_desc("Verbose"), cl::init(false));
 
+
+static cl::opt<bool> VfclibBlackList("vfclibblack-list",
```
pablooliveira: could you  please document black-list, function-file, bitmask, report

src/libvfcreport/libVFCReport.cpp
```diff
@@ -0,0 +1,245 @@
+/********************************************************************************
+ *                                                                              *
+ *  This file is part of Verificarlo.                                           *
+ *                                                                              *
+ *  Copyright (c) 2015                                                          *
```
pablooliveira: 2017

src/libvfcreport/libVFCReport.cpp
```diff
@@ -0,0 +1,245 @@
...
+      writeLineReport(M);
+      for (Module::iterator F = M.begin(), E = M.end(); F != E; ++F) {
+	  writeLineReport(M, *F);
+	  for (Function::iterator BB = F->begin(), BE = F->end(); BB != BE; ++BB) {
+	    writeLineReport(M, *F, *BB);
+	  }
+      }	
+    }     
+   
+    // Ftype getOpType(Instruction &I) {
```
pablooliveira: Remove

tests/test_backends/test_bitmask/bitmask_const.h
```diff
@@ -0,0 +1,40 @@
+#ifndef __BITMASK_CONST_H__
```
pablooliveira: Could you please move this tests to tests/test_bitmask

titeup:
hello, 
Comment again: 
- RAND is only RR and not centered on the real value but on its truncation +0.5. Both can be easily addressed in future version but the current one need to be documented
-RAND is shifting only by one byte the index in the bitfield which means that most of the bits are reused with the next FP operation, I did not notice any example of changing it in my example, but I didn't notice a good sensitivity to your backend
-speaking about sensitivity, in my example, a deflated CG poisson solver, your backend has much less influence tha quad and mpfr even in their RR mode. I also notice that even with the mode selection bug corrected (i checked that rand is correctly used), there is no difference between the mask backend (ZERO, INV, MASK) for a given precision 
- the report path is a compiler report path, i.e. report the static number of operation instrumented. This should be clearly state in the doc. A dynamic version counting "a la gprof" the number of executed FP op per loop per function etc would be quite easy to develop, maybe a future feature?
- You mention me the bug in the mca mode selection, just wanted to mention it in this thread in case someone else falls into it. can you quickly fix it and commit?

On the performance side, based on  real use case measurement, QUAD is 4.2 times faster than MPR on my example, but you are 58 times faster than QUAD. If we can make it work closer to th QUAD/MPFR backend with not loosing too much on performance, your RAND backend is definitly an asset!

titeup:
Hi again,

the report is not working for mee, I checked with verbose, with and without filtering the list of instrumented function and get a correct output. 
When I use the report option I get a csv fil with only the columns name.
Can reproduce it friday when I visit you at UVSQ.
Cheers,

titeup:
any progress?

yohanchatelain:
This is comitted in interflop/interflop fba2aa24ae83531db95365ec34c708cf5323f142.
I just discussed with Pablo and we think it is better to avoid the duplicate merge. Therefore I am closing this PR.

PR #50, opened 2017-05-03
Verificarlo must be initialized before C++ constructors

### LINKED ISSUE #44: Some troubles with a small example using the Boost C++ library
Hi,

First of all thank you for sharing Verificarlo, that is really a promising project!

I just successfully installed it using your debian package

I also run with success the tests (test_tchebychev, test_...) that come with the source. 

I just tried something more complicated, a special function of the Boost C++ library, but I got some troubles...

I wanted to test the  [Boost digamma function](http://www.boost.org/doc/libs/1_63_0/libs/math/doc/html/math_toolkit/sf_gamma/digamma.html)

My small C++ code is

```
#include <boost/math/special_functions/digamma.hpp>
#include <iostream>

int
main()
{
  std::cout << "\n result " << boost::math::digamma(0.1);
}
```
I try to compile it with:

`verificarlo oneExample.cpp -o oneExample`

But I get the following error messages:

```
Unsupported operand type: x86_fp80
Unsupported operand type: x86_fp80
Unsupported operand type: x86_fp80
Unsupported operand type: x86_fp80
Unsupported operand type: x86_fp80
Unsupported operand type: x86_fp80
Unsupported operand type: x86_fp80
Unsupported operand type: x86_fp80
Call parameter type does not match function signature!
x86_fp80 0xK3FFF8000000000000000
 float  %11 = call float %10(x86_fp80 0xK3FFF8000000000000000, x86_fp80 %9)
Stored value type does not match pointer operand type!
  store float %11, x86_fp80* %2, align 16
 x86_fp80Call parameter type does not match function signature!
  %12 = load x86_fp80, x86_fp80* %2, align 16
 float  %16 = call float %15(x86_fp80 %12, x86_fp80 %14)
Stored value type does not match pointer operand type!
  store float %16, x86_fp80* %remainder, align 16
...
 x86_fp80LLVM ERROR: Broken function found, compilation aborted!
/usr/bin/verificarlo: command failed:
/usr/lib/llvm-3.7/bin/opt -S  -load /usr/lib/x86_64-linux-gnu/libvfcinstrument.so -vfclibinst   oneExample.1.ll -o oneExample.2.ll
```

My question, as a new  Verificarlo user, is:
1/ Do I missed something (in my compile command etc...)?
or
2/ Verificarlo is not yet fully fonctional and I have to wait until it supports libs like Boost?
or
3/ Is it a Verificarlo bug?

- Vincent

ps: I also posted this question in the [google group](https://groups.google.com/forum/#!topic/verificarlo/axfHbmUsrMo) but maybe here is a better place.

pablooliveira:
In some boost codes, C++ constructors use floating point arithmetic
and are called before Verificarlo initialises its instrumentation
vtable; causing a SEGFAULT.

By using high priority in the library constructor we can ensure that
verificarlo is always initialized first.

Thanks to Vincent Picaud for reporting and helping debug the issue.

Fixes #44

### DIFF
--- a/src/libvfcinstrument/libVFCInstrument.cpp
+++ b/src/libvfcinstrument/libVFCInstrument.cpp
@@ -125,8 +125,6 @@ namespace {
 
             mca_interface_type = getMCAInterfaceType();
 
-            StringRef SelectedFunction = StringRef(VfclibInstFunction);
-
             // Find the list of functions to instrument
             // Instrumentation adds stubs to mcalib function which we
             // never want to instrument.  Therefore it is important to
--- a/src/vfcwrapper/vfcwrapper.c
+++ b/src/vfcwrapper/vfcwrapper.c
@@ -94,7 +94,7 @@ int vfc_set_precision_and_mode(unsigned int precision, int mode) {
 }
 
 /* vfc_init is run when loading vfcwrapper and initializes vfc libraries */
-__attribute__((constructor))
+__attribute__((constructor(0)))
 static void vfc_init (void)
 {
     char * endptr;


pablooliveira:
We can guarantee that verificarlo is the first by using a particular ld script and forcing a particular order for the constructors. But IMO it's not worth it, doing constructor with priority 0 it' a very rare use case and we should not have conflicts except with very specialized libraries that probably will not perform FP ops in the constructor.

PR #51, opened 2017-06-06
Fix bug on assert enabled llvm version

pablooliveira:
ReplaceInstWithInst checks that the inserted instruction is parent less. This was causing bugs with source-compiled LLVM versions.


pablooliveira:
@titeup, could you please merge this ?

PR #52, opened 2017-06-08
Fix removeInstruction patch

pablooliveira:
Fix issue #51 for scalar case too.


pablooliveira:
@titeup, this patch should fix yesterday install problem from Patrick. Could you please merge it and ask Patrick to try again with the new version ?

Thanks !

titeup:
it worked for Patrick.

PR #54, opened 2017-09-18
Add a test for the Kahan 2x2 linear system example in the paper

pablooliveira:
To avoid introducing a dependency to blas; here we solve the system with the simple Cramer method. Results are consistent with what we show in the paper.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,7 @@ before_install:
 
 install:
   - ./autogen.sh
-  - ./configure --with-dragonegg=/usr/lib/gcc/x86_64-linux-gnu/4.6/plugin/dragonegg.so CC=gcc-4.6 || cat config.log
+  - ./configure --with-llvm=/usr/lib/llvm-3.4/bin/ --with-dragonegg=/usr/lib/gcc/x86_64-linux-gnu/4.6/plugin/dragonegg.so CC=gcc-4.6 || cat config.log
   - make
   - sudo make install
   - export LD_LIBRARY_PATH=/usr/local/lib:$LD_LIBRARY_PATH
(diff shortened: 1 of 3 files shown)


pablooliveira:
Yes I went for something simple. The point is not to compute the determinant; but to reproduce our paper results with a simple test.

> log10(mean(d$x)/sd(d$x))
[1] 8.591463

As a small R test shows, we get the same s value than in the paper with this simple example.

.travis.yml
```diff
@@ -6,11 +6,11 @@ before_install:
   - wget -O - http://llvm.org/apt/llvm-snapshot.gpg.key|sudo apt-key add -
   - sudo add-apt-repository ppa:dns/gnu -y
   - sudo apt-get update
-  - sudo apt-get install libmpfr-dev clang-3.4 llvm-3.4 llvm-3.4-dev dragonegg gcc-4.6 gfortran-4.6 autoconf automake build-essential
+  - sudo apt-get install libmpfr-dev clang-3.4 llvm-3.4 llvm-3.4-dev dragonegg-4.7 gcc-4.7 gfortran-4.7 autoconf automake build-essential
 
 install:
   - ./autogen.sh
-  - ./configure --with-dragonegg=/usr/lib/gcc/x86_64-linux-gnu/4.6/plugin/dragonegg.so CC=gcc-4.6 || cat config.log
+  - ./configure --with-llvm=/usr/lib/llvm-3.4/bin/ --with-dragonegg=/usr/lib/gcc/x86_64-linux-gnu/4.7/plugin/dragonegg.so CC=gcc-4.7 || cat config.log
```
yohanchatelain: Travis brokes at configure since it does not find clang-3.4, maybe change --with-llvm=/usr/lib/llvm-3.4/bin/ by --with-llvm=/usr/local/lib/llvm-3.4/bin/

PR #61, opened 2018-04-06
Remove the offset of 1 on the exponent

yohanchatelain:
It shifts the magnitude of the introduced noise 

        * Added by commit eba90ee2
        * MCA has the property round(inaccurate(x))=x in RandomRounding mode for exact x with t=53 for double (t=24 for single)
        * Add a test (test_rr_mode_exact) to check that property holds.


pablooliveira:
Seems good. Nice idea adding a test :-)

PR #79, opened 2018-11-13
Fix exact operations in RR mode

pablooliveira:
In RR mode we want to avoid adding noise to exact operations. This can be easily done by checking that the result of the operation is exactly representable in the given virtual precision. If the extra precision bits (over the virtual precision) are exactly zero, then the result is exact and no noise is added.

Previously, this property held in RR mode 53 for doubles and 24 for floats when the result was not a power of two. Thanks to D. Defour for the bug report on power of two.

We also revert the offset of 1 change in 5e92113.


pablooliveira:
@yohanchatelain, @titeup could you please review this PR ? Please do not merge yet, as I still want to add some small changes before merging.

Thanks !

PR #82, opened 2019-04-10
Add Branch support

ddavid42:


titeup:
It would be could to add a test for this feature in future release, I will open an issue for it.

pablooliveira:
@titeup, @ddavid42 thanks for this commit.
The feature is good but the history shown by git log is not very nice or clear.

A few minor recommendations to keep the project git history clean next time:
* when committing please set correctly the author name (all the commits in this PR are attributed to gitolite <gitolite@tahiti.prism.uvsq.fr> instead of @ddavid42)
* when merging, if there are various commits with the same description "add test support", please use a squash merge instead and write a full description of the work merged.

Thanks !

PR #86, opened 2019-08-05
Fix shell expansion in verificarlo

pablooliveira:
Verificarlo passes extra options to clang through the shell, if the arguments are shell expanded this can cause problems.

Example: 
   ```verificarlo -DPLATFORM="linux"``` 
gets expanded to
   ```verificarlo -DPLATFORM=linux``` 
one clang is called which fails.

Reported by @gkiar. 

This issue is a work in progress,
- [x] Does the fix proposed is general enough?
- [x] Must add some tests
- [x] Should we consider rewritting verificarlo.in.in to perform subprocess calls directly without going through the shell? This could be more robust, but may bring also problems with PATH and environment and requires rewritting a good part of the wrapper.

### DIFF
--- a/src/libmca-quad/mcalib.c
+++ b/src/libmca-quad/mcalib.c
@@ -380,40 +380,35 @@ static inline double _mca_dbin(double a, double b, const int qop) {
 * point operators
 **********************************************************************/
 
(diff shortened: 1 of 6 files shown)


pablooliveira:
After some though, escaping options so they are not expanded during the `shell` call seems robust. I added a simple test reproducing this issue to our test suite.

PR #88, opened 2019-08-06
Implement exclude support

pablooliveira:
This PR implements a flexible inclusion/exclusion mechanism.

Thanks to @gkiar for providing an interesting use-case that requires inclusion/exclusion support.

## Inclusion / Exclusion

If you only wish to instrument a specific function in your program, use the
`--function` option:

```bash
   $ verificarlo *.c -o ./program --function=specificfunction
```

For more complex scenarios, a white-list / black-list mechanism is also available
through the options `--include-file INCLUSION-FILE` and `--exclude-file EXCLUSION-FILE`.

`INCLUSION-FILE` and `EXCLUSION-FILE` are files specifying which modules and
functions should be included or excluded from Verificarlo instrumentation.
Each line has a module name followed by a function name. Both the module or
function name can be replaced by the wildcard `*`. Empty lines or lines
starting with `#` are ignored.

```
# include.txt
# this inclusion file will instrument f1 in main.c and util.c, and instrument
# f2 in util.c everything else will be excluded.
main f1
util f1
util f2


# exclude.txt
# this exclusion file will exclude f3 from all modules and all functions in
# module3.c
* f3
module3 *
```

Inclusion and exclusion files can be used together, in that case inclusion
takes precedence over inclusion.

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,13 +117,6 @@ If you are trying to compile a shared library, such as those built by the Cython
 extension to Python, you can then also set the shared linker environment variable
 (`LDSHARED='verificarlo -shared'`) to enable position-independent linking.
 
-If you only wish to instrument a specific function in your program, use the
-`--function` option:
-
-```bash
-   $ verificarlo *.c -o ./program --function=specificfunction
-```
-
 When invoked with the `--verbose` flag, verificarlo provides detailed output of
 the instrumentation process.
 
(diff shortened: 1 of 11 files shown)


gkiar:
This looks great, @pablooliveira ! I'm trying it out now :) A bit pre-emptively, if I want to use this with `Makefile` builds, would I set my compiler/linker variables to be something like the following?

```bash
$ export CC="verificarlo --exclude-file=exclude.txt"
```

pablooliveira:
@gkiar, indeed, `$ export CC="verificarlo --exclude-file=exclude.txt"` would work.

If you look at the proposed changes on the Dockerfile at gkiar/stability-mca#2, that is the way the Dockerfile passes the exclusion file to Verificarlo.

An example of exclude file for python build is [also available](https://github.com/gkiar/stability-mca/blob/8fa39ad10b348fa069e154a9f74af99951c002e9/code/verificarlo/fuzzy-python/python-vfc-exclude.txt).

PR #93, opened 2019-08-07
Move towards interflop

pablooliveira:
As discussed with @titeup and @yohanchatelain, we would like to move verificarlo master so it uses the new interflop interface.
This presents advantages in terms of interoperability with other tools. It also should result in a cleaner separation of concerns between libvfcinstrument and the backends.

This PR is a work in progress please do not merge yet.

- [x] Change mca and mca-mpfr backends to use a longopt sys.argv parser to get its arguments
- [x] Document all changes in README.md
- [ ] Update tutorial


pablooliveira:
Since this is a big change and requires changes in the documentation. It is important we release a new version of verificarlo before merging this into master.

pablooliveira:
@yohanchatelain 917ac74 implements the check that you suggested during the initial review.

PR #97, opened 2019-08-13
Fix #96: fix support for bool vectors

### LINKED ISSUE #96: bool vectors are forbidden in LLVM 4.0
Commit #84 breaks support for LLVM 4.0 since extended [bool vectors are forbidden](https://lists.llvm.org/pipermail/cfe-commits/Week-of-Mon-20160125/148245.html). Reported by @yohanchatelain, Thanks !

pablooliveira:
Bool vectors are not supported in LLVM, the support for fcmp vector
instructions was broken and we could not return the proper value.

Here we rewrite fcmp vector hooks to return int vectors instead,
libVFCInstrument is modified to perform an appropriate cast when
replacing vector fcmp instructions.

We add a runtime test that checks that the return value of fcmp
instructions is properly handled.

Thanks to @yohanchatelain for reporting the bug.


pablooliveira:
@yohanchatelain could you please check if this fixes #96 in llvm 4.0 and merge ?
Thanks!

PR #105, opened 2019-08-23
Reset LIBS dependencies in configure.ac

pablooliveira:
Reset libs after ac_check_lib. We do not want each and every backend or the
libvfcinstrument frontend to be linked against libmpfr or libgfortran for
example. -l flags should be added individually in each Makefile.am

* Fix a bug with static link mode.
* Added a test for static link mode.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -57,6 +57,11 @@ EOF
     fi
 fi
 
+# Reset libs after ac_check_lib We do not want each and every backend or the
+# libvfcinstrument frontend to be linked against libmpfr or libgfortran for
+# example. -l flags should be added individually in each Makefile.am
+LIBS=""
+
 # Check for header files.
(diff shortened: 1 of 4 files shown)


src/backends/interflop-mca/Makefile.am
```diff
@@ -1,5 +1,5 @@
 lib_LTLIBRARIES = libinterflop_mca.la
 libinterflop_mca_la_SOURCES = interflop_mca.c
-libinterflop_mca_la_LDFLAGS = -lm
+libinterflop_mca_la_LDFLAGS = -lm -lquadmath
```
yohanchatelain: You should remove -lquadmath and add a test for checking if `__float128` is supported by the compiler.
It removes dependency and gets the configure.ac cleaner.

pablooliveira: Sure. I removed the dependency and check in configure.ac for now.
Checking for __float128 is not trivial, I will open an issue and move it to another PR.

pablooliveira: Opened issue #110

PR #108, opened 2019-08-26
Delta debug

pablooliveira:
This PR will track progress in implementing delta-debug for verificarlo.

* A first implementation is working.
* Tests have been added for,
  * The rectangle integration (taken from verrou's tutorial)
  * Simple generation / inclusion tests to validate the address filtering logic in `vfcwrapper/main.c`

Ongoing steps:

- [ ] Update documentation: add a section on delta-debug usage and update dependencies
- [ ] Check that the docker image works
- [x] Improve error messages in DD_stoch if needed


pablooliveira:
@lathuili, thanks for your excellent help integrating your DD.py version into verificarlo! The abstraction provided by `DD_stoch` worked well, I only had to change `dd_line.py` to make it work with verificarlo. I also changed some env variables / user messages so they use the INTERFLOP prefix. 

I borrowed verrou's rectangle integration test-case as a first integration test for this feature and tests are passing !

If you have any comments or feedback feel free to ping me. 

Since now parts of our repositories (mca quad backend and DD.py) are shared between verrou and verificarlo; it would be nice to discuss how to maintain them together. Ideally, I think we could use the interflop repository as a common repository to commit changes, which we can then import in our respective projects either as a submodule or manually.

pablooliveira:
@titeup could you please review this PR ? I would like to merge it today to prepare the docker image for next week tutorial. The documentation is still missing but will be pushed later.
Thanks :smile:

PR #112, opened 2019-08-27
Fixes #111

### LINKED ISSUE #111: Issue in the qnoise function
https://github.com/verificarlo/verificarlo/blob/6b86bcd044edbe316a3c766a84f5de5cb2603ea3/src/backends/interflop-mca/interflop_mca.c#L172-L173
The `qnoise` function computes `2^(exp)*d_rand` so if the exponent is 0 we should return `d_rand` instead of 1.

https://github.com/verificarlo/verificarlo/blob/6b86bcd044edbe316a3c766a84f5de5cb2603ea3/src/backends/interflop-mca/interflop_mca.c#L178-L179

yohanchatelain:
Fixes #111


pablooliveira:
Could you actually remove the whole if block ? Not sure if we need to make a special case out of this.

PR #115, opened 2019-08-29
Add seed fixing mechanism

yohanchatelain:
This PR allows to select a fixed seed for the RNG in mca backends.

### DIFF
--- a/src/backends/interflop-mca-mpfr/interflop_mca_mpfr.c
+++ b/src/backends/interflop-mca-mpfr/interflop_mca_mpfr.c
@@ -52,6 +52,11 @@
 #include "../../common/interflop.h"
 #include "../../common/tinymt64.h"
 
+typedef struct {
+  int choose_seed;
+  int seed;
+} t_context;
+
 /* define the available MCA modes of operation */
 #define MCAMODE_IEEE 0
 #define MCAMODE_MCA 1
@@ -152,17 +157,23 @@ static int _mca_inexact(mpfr_ptr a, mpfr_rnd_t rnd_mode) {
   mpfr_clear(mpfr_zero);
 }
 
-static void _mca_seed(void) {
+static void _set_mca_seed(int choose_seed, int seed) {
   const int key_length = 3;
   uint64_t init_key[key_length];
-  struct timeval t1;
-  gettimeofday(&t1, NULL);
-
-  /* Hopefully the following seed is good enough for Montercarlo */
-  init_key[0] = t1.tv_sec;
-  init_key[1] = t1.tv_usec;
-  init_key[2] = getpid();
-
+  
+  if (choose_seed) {
+    init_key[0] = seed;
+    init_key[1] = seed;
+    init_key[2] = seed;    
+  } else {    
+    struct timeval t1;
+    gettimeofday(&t1, NULL);
+    /* Hopefully the following seed is good enough for Montercarlo */
+    init_key[0] = t1.tv_sec;
+    init_key[1] = t1.tv_usec;
+    init_key[2] = getpid();
+  }
+  
   tinymt64_init_by_array(&random_state, init_key, key_length);
 }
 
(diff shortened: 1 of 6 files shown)


pablooliveira:
Please rebase this on current master and remove first commit. If not it will appear twice in the history with two different sha sums. You can do a rebase followed by a force push.

src/backends/interflop-mca-mpfr/interflop_mca_mpfr.c
```diff
@@ -152,17 +157,23 @@ static int _mca_inexact(mpfr_ptr a, mpfr_rnd_t rnd_mode) {
...
-  /* Hopefully the following seed is good enough for Montercarlo */
-  init_key[0] = t1.tv_sec;
-  init_key[1] = t1.tv_usec;
-  init_key[2] = getpid();
-
+  
+  if (choose_seed) {
+    init_key[0] = seed;
+    init_key[1] = seed;
+    init_key[2] = seed;
```
pablooliveira: instead of doing this and then calling `tinymt64_init_by_array` you could directly call `void tinymt64_init(tinymt64_t * random, uint64_t seed);` which would be cleaner. The seed should be of type uint64_t.

src/backends/interflop-mca/interflop_mca.c
```diff
@@ -306,17 +311,23 @@ static int _mca_inexactd(double *da) {
...
-  /* Hopefully the following seed is good enough for Montercarlo */
-  init_key[0] = t1.tv_sec;
-  init_key[1] = t1.tv_usec;
-  init_key[2] = getpid();
-
+  
+  if (choose_seed) {
+    init_key[0] = seed;
+    init_key[1] = seed;
+    init_key[2] = seed;
```
pablooliveira: same remark than above

src/backends/interflop-mca/interflop_mca.c
```diff
@@ -464,6 +477,16 @@ static error_t parse_opt(int key, char *arg, struct argp_state *state) {
         errx(1, "interflop_mca: --mode invalid value provided, must be one of: {ieee, mca, pb, rr}.");
       }
       break;
+    case 's':
+      errno = 0;
+      int seed = strtol(arg, &endptr, 10);
```
pablooliveira: same remark than above

tests/test_fix_seed/test.sh
```diff
@@ -0,0 +1,57 @@
...
+
+	export VFC_BACKENDS="${BACKENDSO} --precision $PREC --mode $MODE --seed=0"
+	./test > out_seed_1
+	./test > out_seed_2
+	Check_similarity out_seed_1 out_seed_2
+	    
+    done
+}
+
+for op in "+" ; do
```
pablooliveira: Since we are only checking for op +, could we simplify the test ? test.c seems overly complicated for what we are testing.
Just a simple iteration that accumulates 100 times the value .01 should be enough.

src/backends/interflop-mca-mpfr/interflop_mca_mpfr.c
```diff
@@ -329,7 +342,17 @@ static error_t parse_opt(int key, char *arg, struct argp_state *state) {
       } else if (strcasecmp(MCAMODE[MCAMODE_RR], arg) == 0) {
         _set_mca_mode(MCAMODE_RR);
       } else {
-        errx(1, "interflop_mca: --mode invalid value provided, must be one of: {ieee, mca, pb, rr}.");
+        errx(1, "interflop_mca_mpfr: --mode invalid value provided, must be one of: {ieee, mca, pb, rr}.");
+      }
+      break;
+    case 's':
+      errno = 0;
+      int seed = strtol(arg, &endptr, 10);
```
pablooliveira: no need for the temporary seed. you can directly write inside `ctx-> seed` here and make the code a bit shorter.

PR #122, opened 2019-10-17
added explanation for fixing python path error

mistoan:


README.md
```diff
@@ -102,6 +102,18 @@ install procedure:
    $ make installcheck
 ```
 
+In order to take use the delta debug features, you need to export the path of the corresponding python packages. For example, for a global install, this would resemble (edit for your installed Python version):
```
pablooliveira: - In order to take use -> In order to use

pablooliveira: - This should come before `make installcheck`, because tests will fail if PATH is not exported.

README.md
```diff
@@ -102,6 +102,18 @@ install procedure:
...
+
+```bash
+export PYTHONPATH=${PYTHONPATH}:/usr/local/lib/pythonXXX.XXX/site-packages
+```
+
+Alternatively, you can make these changes permanent by editing your ./bashrc file by adding the above line, and then reloading your environment using:
```
pablooliveira: - ./bashrc -> `~/.bashrc`, `~/.profile` or relevant shell configuration file,

PR #124, opened 2019-10-30
Add binary output option

ddavid42:
Modifications to add binary output support. 
This output is useful to have a fine analysis of what is happening at every bit.

### DIFF
--- a/src/backends/interflop-ieee/interflop_ieee.c
+++ b/src/backends/interflop-ieee/interflop_ieee.c
@@ -14,70 +14,6 @@ typedef struct {
 
 #define STRING_MAX 256
 
-void float_to_binary(float f, char* s_val){
- union ieee754_float ud;
- unsigned char sign_field;
- unsigned short exponent_field;
- short exponent;
- unsigned long long fraction_field, significand;
- int i, str_i, start = 0, end = 23;
- char s_exp[80];
-
- ud.f = f;
- sign_field = ud.ieee.negative;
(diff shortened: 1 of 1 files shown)


pablooliveira:
@ddavid42: Thanks for this, looks good!

In the interest of keeping the code short, there is a more succint way to convert a float/double to its binary representation: take a pointer to the float, and reinterpret_cast it to a pointer to char. Example here in C++ (but it also works in C):

http://www.cplusplus.com/forum/general/60160/

Would you consider using that instead ? I think it would make the code much shorter and maintainable.

ddavid42:
No, as this code does not handle the binary value but the binary representation instead.

pablooliveira:
I'm not sure to understand the difference. Could you explain your point ?

ddavid42:
Binary representation of value 1.75 is:
00111111111000000000000000000000
Whereas Binary value is:
1.11 2^0

pablooliveira:
Ok I see, this requires indeed to parse excedent field and sign and have special cases (denormals, etc).

I see why your code is longer and requires the added complexity. We might have used %a printf specifier; but that would have given us hexadecimal.

Do you think we could factor somehow the double and single conversion functions ? There seems to be some repetition there ? 

Apart from the added code complexity, I think this is a great addition when debugging ! Hopefully we can trimmer it a bit down, but I'm all for merging it !

ddavid42:
Yes sure there are some possible factorisations. I'll commit them.

PR #130, opened 2019-11-22
Fix documentation about fortran tests

KillianBabilotte:
Fix documentation: fortran tests are skipped when dragonegg support is missing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,7 +83,7 @@ verificarlo works as expected on your system:
    $ make installcheck
 ```
 
-If you disable dragonegg support during configure, fortran_test will be disable and be consider as passing the test.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -83,7 +83,7 @@ verificarlo works as expected on your system:
    $ make installcheck
 ```
 
-If you disable dragonegg support during configure, fortran_test will fail.
+If you disable dragonegg support during configure, fortran_test will be disable and be consider as passing the test.
```
pablooliveira: could you please change to "be disable**d** and considered as passing the test."

KillianBabilotte: of course ! Sorry...

Issue #131, opened 2019-11-26
bug with argument parsing in the backends


DamDem:
[test.log](https://github.com/verificarlo/verificarlo/files/3891699/test.log)
[config.log](https://github.com/verificarlo/verificarlo/files/3891700/config.log)
I installed verificarlo on my device following the README.md.  After running "make installcheck"  test_ieee_backend failed. I linked the test.log related on this test and the config.log.

pablooliveira:
@DamDem thanks for the report !

@yohanchatelain you mentioned that you also got this bug and knew how to fix it; I'll assign it to you. Thanks :muscle:

PR #132, opened 2020-01-17
Fix issue with argument parsing in ieee backend

yohanchatelain:
Closes issue #131.
According to the [man page](https://www.gnu.org/software/libc/manual/html_node/Argp-Example-3.html#Argp-Example-3), a `return 0` was missing:

> It should return either 0, meaning success, ARGP_ERR_UNKNOWN, meaning the given KEY wasn’t recognized, or an errno value indicating some other error.


pablooliveira:
:tada: This looks good, merged !

PR #133, opened 2020-01-22
fix coverity report

titeup:
The only unfixed bug was in llvm3.3 source (missing struct initialization)
Since this is out of our field and in an older version of LLVM I haven't fix it.
Coverity report was a great tool, I advise to use it on a more regular bases on our project.


yohanchatelain:
We could plan to add coverity in the github pipeline.

pablooliveira:
@yohanchatelain go ahead with coverity integration in github pipeline, excellent idea !

@titeup congrats for the fix, thanks !

PR #139, opened 2020-02-03
Update MCA backends

### LINKED ISSUE #104: Add guard to avoid spurious precisions
https://github.com/verificarlo/verificarlo/blob/be678156e1f50009c2e01a04d5efd4dacf53321f/src/backends/interflop-mca-mpfr/interflop_mca_mpfr.c#L97-L99
https://github.com/verificarlo/verificarlo/blob/be678156e1f50009c2e01a04d5efd4dacf53321f/src/backends/interflop-mca/interflop_mca.c#L102-L105

### LINKED ISSUE #106: lib_mca backends could have a different virtual precision for floats and doubles
We could set a different virtual precision for floats and doubles in mca backends. This would allow to run for example RR/53 for doubles and RR/24 for floats. This change is pretty simple and [has already been added](https://github.com/edf-hpc/verrou/blob/bf6eca9d31e6a6a61150c90998f52c1d7bdf9c95/backend_mcaquad/mcalib.c#L91) by @lathuili, we should backport his change here in both libinterflop_mca and libinterflop_mca_mpfr.

### LINKED ISSUE #110: configure.ac should check for __float128
libinterflop-mca backend (quad backend) requires support for __float128.

We should check that this type is supported in configure.ac. 

https://gitlab.freedesktop.org/pixman/pixman/commit/09600ae7e34eb777471c931cd4c3a8cdbda6e84a gives an example of how to do this.

yohanchatelain:
- interlop-mca & interflop-mca-mpfr:
	 * Adds separate virtual precision for binary32 and binary64.
	   Available through options --precision-binary32 and --precision-binary64.
	   Closes #106.
	 * Adds guards to avoid spurious precisions that closes #104.
	 * Uses `enum` for MCA_MODE instead of #49 
     * Renames _mca_{s,d}bin functions into _mca_binary{32,64}_binary_op.
- interflop-mca:
       	 
     * Uses binary128 structure and bitfields to handle __float128 and
       	   enhances the readabilty of the code.
- interflop-mca-mpfr:
     * For avoiding duplication of code dues to separated virtual precisions,
	   uses a factored version of _MCA_INEXACT as a macro function.
	   This version uses the `_Generic` feature of C++11 available since
	   gcc-4.9.
- tests:
     * Updated for using options --precision-binary{32,64}
- configure.ac:
    * Tests that __unint128_t and  __float128_t are supported.
	   This closes #110 
    * Tests that _Generic keyword is supported by gcc.
	   Used in interflop-mca-mpfr backend.
    * Updated with missing checks reported by autoscan.

- .travis.yml & Dockerfile:
    * Updated for testing with gcc-4.9.


Dockerfile
```diff
@@ -3,29 +3,48 @@
...
 MAINTAINER Pablo Oliveira <pablo.oliveira@uvsq.fr>
 
+ENV LLVM_VERSION 3.9
+ENV GCC_VERSION 5
+ENV GCC_PATH /usr/lib/gcc/x86_64-linux-gnu/${GCC_VERSION}
+
 # Retrieve dependencies
-RUN apt-get -y update && apt-get -y install --no-install-recommends bash ca-certificates make git libmpfr-dev clang-3.5 llvm-3.5 llvm-3.5-dev dragonegg-4.7 gcc-4.7 g++-4.7 gfortran-4.7 libgfortran-4.7-dev autoconf automake libedit-dev libtool libz-dev python3 python3-numpy python3-matplotlib binutils vim && rm -rf /var/lib/apt/lists/
-ENV LIBRARY_PATH /usr/lib/gcc/x86_64-linux-gnu/4.7:/usr/lib/llvm-3.5/lib:$LIBRARY_PATH
+# RUN apt-get -y update && apt-get -y install --no-install-recommends software-properties-common
```
pablooliveira: Please remove these commented lines

Dockerfile
```diff
@@ -3,29 +3,48 @@
...
+  ./configure --with-llvm=$(llvm-config-${LLVM_VERSION} --prefix) --with-dragonegg=${DRAGONEGG_PATH} CC=${GCC_VERSION} CXX=g++-{GCC_VERSION}
 
 # Build and test verificarlo
 
 ENV LD_LIBRARY_PATH /usr/local/lib:$LD_LIBRARY_PATH
 ENV PATH /usr/local/bin:$PATH
 ENV PYTHONPATH /usr/local/lib/python3.4/site-packages/:$PYTHONPATH
 RUN cd verificarlo && make && make install && make installcheck
-
-# Setup working directory
```
pablooliveira: Why should we remove these ? Is the tutorial still working with this image, if we make this change ?

src/backends/interflop-mca/interflop_mca.c
```diff
@@ -122,135 +152,85 @@ static double _mca_rand(void) {
...
-  // do not reuse it outside this code!
-  uint64_t hx, ix;
-  uint32_t exp = 0;
-  GET_FLT128_MSW64(hx, x);
-  // remove sign bit, mantissa will be erased by the next shift
-  ix = hx & QUAD_HX_ERASE_SIGN;
-  // shift exponent to have LSB on position 0 and complement
-  exp = (ix >> QUAD_HX_PMAN_SIZE) - QUAD_EXP_COMP;
-  return exp;
+static inline int32_t rexpd(double d) {
```
pablooliveira: Could you please briefly comment what each function does, to make code more readable please ?

src/backends/interflop-mca/interflop_mca.c
```diff
@@ -54,30 +54,39 @@
...
+  bool choose_seed;
   uint64_t seed;
 } t_context;
 
 /* define the available MCA modes of operation */
-#define MCAMODE_IEEE 0
-#define MCAMODE_MCA 1
-#define MCAMODE_PB 2
-#define MCAMODE_RR 3
+typedef enum { ieee, mca, pb, rr } mcamode;
```
titeup: IMHO I found the older names more readable (MCAMODE_xx) I don't care about using enum or define variable, but it would be easier to read to explicitly expose that we are talking about MCAMODE and not about some local variable with lowercase names.

src/backends/interflop-mca/interflop_mca.c
```diff
@@ -122,135 +152,85 @@ static double _mca_rand(void) {
...
-    *x = DOUBLE_PLUS_INF;
-    res = *((double *)x);
-    return res;
-  }
-  if (exp < -1022) { /*subnormal*/
-    *x = ((uint64_t)DOUBLE_PMAN_MSB) >> -(exp + DOUBLE_EXP_MAX);
-    res = *((double *)x);
-    return res;
-  }
+static inline double pow2d(int exp) { return ldexp(1.0, exp); }
```
titeup: This is old. But I kind of remember we didn't use existing function because our very specific case allows for some short cuts and performance gain. Did you check the parformance of your backend compare to the original quad one?

Dockerfile
yohanchatelain: Done.

Dockerfile
yohanchatelain: It was a mistake. Thank you for pointing out.

src/backends/interflop-mca/interflop_mca.c
yohanchatelain: Done.

src/backends/interflop-mca/interflop_mca.c
```diff
@@ -122,135 +152,85 @@ static double _mca_rand(void) {
...
+  /* we return an sign(rand) * Infinity */
+  else if (noise_exp > QUAD_EXP_MAX) {
+    noise.ieee128.sign = brand.ieee.sign;
+    noise.ieee128.exponent = QUAD_EXP_MAX;
+    noise.ieee128.mantissa = 0;
+  }
+  /* We are in the subnormal case */
+  else if (noise_exp < -QUAD_EXP_MIN) {
+    /* test if result is below the minimal subnormal quad number */
+    if (noise_exp < -(QUAD_EXP_MIN + QUAD_PMAN_SIZE)) {
```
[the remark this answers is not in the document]
yohanchatelain: Actually, we are never to be going in this case since the exponent value comes from a double and will never reach the QUAD_EXP_MIN bound. Moreover, your original code was wrong since this test tests if we are in the subnormal case and you set to -infinity after that.

src/backends/interflop-mca/interflop_mca.c
```diff
@@ -272,11 +252,12 @@ static bool _is_representabled(double *da) {
    * in the current virtual precision */
   uint64_t p_mantissa = (*((uint64_t *)da)) & DOUBLE_GET_PMAN;
   /* here we know that (MCALIB_T-1) < 53 */
-  return ((p_mantissa << (MCALIB_T + DOUBLE_EXP_SIZE)) == 0);
+  return ((p_mantissa << (MCALIB_BINARY32_T + DOUBLE_EXP_SIZE)) == 0);
 }
 
 static int _mca_inexactq(__float128 *qa) {
```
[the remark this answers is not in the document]
yohanchatelain: I'm not sure to see which function you're talking about since is_representabled returns a `bool`

src/backends/interflop-mca/interflop_mca.c
yohanchatelain: I think that `enum` is better since

> Enumerations permit the declaration of named constants in a more convenient and structured fashion than does #define; they are visible in the debugger, obey scope rules, and participate in the type system.

but I agree with the readability.

src/backends/interflop-mca/interflop_mca.c
yohanchatelain: Yes, you're right. Actually, as for `qnoise` we don't need to check specials cases.

src/backends/interflop-mca/interflop_mca.c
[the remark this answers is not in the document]
yohanchatelain: As we talked, we should switch the return type of these functions from `int` to `void` and use `errx` and `warnx ` instead.

PR #144, opened 2020-02-14
Add support for llvm-5.0 to llvm-9

yohanchatelain:


src/libvfcinstrument/libVFCInstrument.cpp
```diff
@@ -43,11 +43,38 @@
...
+ * M.getOrInsertFunction(name,res,,(Type*)NULL) */
+/* It could be fixed when __VA_OPT__ will be available (see
+ * https://gcc.gnu.org/onlinedocs/cpp/Variadic-Macros.html)*/
+#define GET_OR_INSERT_FUNCTION(M, name, res, ...)                              \
+  M.getOrInsertFunction(name, res, __VA_ARGS__, (Type *)NULL)
+typedef llvm::Constant *_LLVMFunctionType;
+#elif LLVM_VERSION_MAJOR < 5
+#define CREATE_CALL3(func, op1, op2, op3)                                      \
+  (Builder.CreateCall(func, {op1, op2, op3}, ""))
+#define CREATE_CALL2(func, op1, op2) (Builder.CreateCall(func, {op1, op2}, ""))
```
pablooliveira: CREATE_CALL2 seems always the same in these #ifdef blocks ?
Should we keep it here ?

src/libvfcinstrument/libVFCInstrument.cpp
```diff
@@ -248,15 +275,14 @@ struct VfclibInst : public ModulePass {
...
+      _LLVMFunctionType hookFunc = GET_OR_INSERT_FUNCTION(
+          M, mcaFunctionName, res, Builder.getInt32Ty(), opType, opType);
       newInst = CREATE_CALL3(hookFunc, Builder.getInt32(FCI->getPredicate()),
                              FCI->getOperand(0), FCI->getOperand(1));
       newInst = Builder.CreateIntCast(newInst, retType, true);
     } else {
-      Constant *hookFunc = M.getOrInsertFunction(mcaFunctionName, retType,
-                                                 opType, opType, (Type *)0);
+      _LLVMFunctionType hookFunc =
+          GET_OR_INSERT_FUNCTION(M, mcaFunctionName, retType, opType, opType);
```
pablooliveira: Why are we passing opType twice to this macro ?

yohanchatelain: Because it is the type of the arguments. Since we have two arguments, we have to pass two `opType`.

src/libvfcinstrument/libVFCInstrument.cpp
yohanchatelain: Yes, it is. I prefer that each version has its definition for readability purposes, even it is the same. But we can factorize it if you prefer.

src/libvfcinstrument/libVFCInstrument.cpp
pablooliveira: Ok sorry, I misread.

src/libvfcinstrument/libVFCInstrument.cpp
pablooliveira: Ok we can keep it like that.

PR #152, opened 2020-02-26
Pre-release 0.3.0

yohanchatelain:
- Add modifications to CHANGELOG
	  * Add new contributions since the last release in the Unreleased section
	  * Add sections Added/Fixed/Removed
	- Put common functions in float_utils.c and options.c
	  * float_utils.c: group functions relative to floating point manipulations
	  * options.c: group common function for parsing functions
	- Add --daz and --ftz options
	  * --daz: denormals-are-zero, flush subnormal inputs to zero
	  * --ftz: flush-to-zero, flush subnormal output to zero
	- Add logger
	  * Three levels: info, warning and error
	  * Wrap fprintf, warnx and errx
	  * Use color to enhance readability
	  * VFC_BACKENDS_LOGGER="False" turns off the logger
	  * VFC_BACKENDS_COLORED_LOGGER="False" turns off the color
	- interflop_mca.so, interflop_mca_mpfr.so and interflop_bitmask.so
	  * Add DAZ/FTZ support
	  * Replace warnx/errx by logger_warning/error
	  * Use variables for options name instead of hardcoded one
	- interflop_mca.so and interflop_bitmask.so
	  * Factorize _inexact functions into _INEXACT macro function
	- interflop_ieee.so
	  * Rename headers interflop_ieee{_bin} into Decimal, Binary
	    since we have the name of the backend displayed with the logger now

PR #153, opened 2020-03-03
WIP: Cancellation Backend

KillianBabilotte:
Work achieved by our team during the first semester of our annual project from our Master degree under @pablooliveira direction.


tests/test_ddebug_canc/test.log
```diff
@@ -0,0 +1,12 @@
+make[4] : on entre dans le répertoire « /home/alan/Bureau/Lib/verificarlo/tests/test_ddebug_canc »
```
pablooliveira: `tests/test_ddebug_canc/test.log` should not be committed since it is generated when the tests are run.

tests/test_ddebug_canc/test.sh
```diff
@@ -0,0 +1,16 @@
...
+if grep "archimedes.c:16" dd.line/rddmin-cmp/dd.line.exclude ; then
+  if grep "archimedes.c:17" dd.line/rddmin-cmp/dd.line.exclude ; then
+    echo "success !"
+    exit 0
+  else
+    echo "missing line 17 (cancellation)"
+    exit 1
+  fi
+else
+  echo "missing line 16 (round-off)"
```
pablooliveira: This test should be changed. When run with the cancellation backend, **only line 17** should be detected.

src/backends/interflop-cancellation/interflop_cancellation.c
```diff
@@ -0,0 +1,280 @@
...
+  /* Get size of the exponent */                      \
+  int _E_ = (6+3*(sizeof(X) >> 2)-1);                 \ 
+  /* Get the size of the mantiss */                   \
+  int _M_ = (sizeof(X)*8-_E_-1);                      \
+  /* Compute the exponent */                          \     
+  (int)bit_a_b(_V_>>_M_,0,_E_-1) - ((1<<_E_-1)-2);})
+
+#define inexact(X,S) (X + pow(2,exp(X)-S) * _mca_rand())
+
+static void _set_mca_seed(int choose_seed, uint64_t seed) {
```
pablooliveira: This will have to be refactored after merging #152

(now there is a standard function in `src/common`)

src/backends/interflop-cancellation/interflop_cancellation.c
```diff
@@ -0,0 +1,280 @@
...
+
+// perform_bin_op: applies the binary operator (op) to (a) and (b)
+// and stores the result in (res)
+
+#define mant(X) ((int)(sizeof(X)*8-(6+3*(sizeof(X) >> 2)-1)-1))
+
+#define cancell(X,Y,Z) ({                                         \
+  int cancellation = detect(X,Y,Z);                               \
+  if(cancellation >= MCALIB_T){                                   \
+    if(WARN){                                                     \
```
pablooliveira: Use the new logger library after merge of #152

src/backends/interflop-cancellation/interflop_cancellation.c
```diff
@@ -0,0 +1,280 @@
...
+      printf("cancellation of size %d detected\n", cancellation);}\
```
pablooliveira: Use clang-format to format the C code here (brace should be on the next line).

src/backends/interflop-cancellation/interflop_cancellation.c
```diff
@@ -0,0 +1,280 @@
...
+  Z;})
+
+/************************* FPHOOKS FUNCTIONS *************************
+ * These functions correspond to those inserted into the source code
+ * during source to source compilation and are replacement to floating
+ * point operators
+ **********************************************************************/
+
+static void _interflop_add_float(float a, float b, float *c, void *context) {
+  *c = a + b;
```
pablooliveira: **Bug: what if a and b** are of different sign !

cancell should then be called.

src/backends/interflop-cancellation/interflop_cancellation.c
```diff
@@ -0,0 +1,280 @@
...
+  *c = a * b;
+}
+
+static void _interflop_div_float(float a, float b, float *c, void *context) {
+  *c = a / b;
+}
+
+static void _interflop_add_double(double a, double b, double *c,
+                                  void *context) {
+  *c = a + b;
```
pablooliveira: Same **bug than before**, a and b can be of different sign. In this case cancell should be called.

src/backends/interflop-cancellation/interflop_cancellation.c
```diff
@@ -0,0 +1,280 @@
...
+  /* Convert X to an int */                           \
+  u_int64_t _V_ = *(u_int64_t*)&X;                    \
+  /* Get size of the exponent */                      \
+  int _E_ = (6+3*(sizeof(X) >> 2)-1);                 \ 
+  /* Get the size of the mantiss */                   \
+  int _M_ = (sizeof(X)*8-_E_-1);                      \
+  /* Compute the exponent */                          \     
+  (int)bit_a_b(_V_>>_M_,0,_E_-1) - ((1<<_E_-1)-2);})
+
+#define inexact(X,S) (X + pow(2,exp(X)-S) * _mca_rand())
```
pablooliveira: This can be optimized. Because noise is only added to cancellation bits (which are necessarily zero); it is enough to generate a rand field, shift it to the first cancelled bit and OR it to the result.

src/backends/interflop-cancellation/quadmath-imp.h
```diff
@@ -0,0 +1,189 @@
+/* GCC Quad-Precision Math Library
```
pablooliveira: This file should not be needed as per above comments.

src/backends/interflop-cancellation/interflop_cancellation.c
pablooliveira: Thinking about this. The advantage of using MCA noise is that it is centered on the IEEE result. The shift and or approach I propose will generate a bias... I'm not sure yet which model is better to simulate effect of cancellations on hardware. We could have both modes; but this requires further study.

tests/test_ddebug_canc/archimedes.c
```diff
@@ -0,0 +1,32 @@
+#include <math.h>
```
yohanchatelain: Binary _archimedes_ should not be included in the PR.

README.md
```diff
@@ -387,6 +388,29 @@ The option `--seed` fixes the random generator seed. It should not
...
+This backend can be used with three options.
+
+The option `--tolerance` sets the tolerance within the backend will trigger a
+cancellation. By default tolerance is set to 1.
+
+The option `--warning` warns on the standard output stream when a cancellation is
+triggered by the backend.
+
+The option `--seed` fixes the random generator seed. It should not generally be
+used except if one to reproduce a particular MCA trace.
```
yohanchatelain: Could you use the same format as for the other backends to present the options?

tests/test_cancellation/test_canc_detect.c
```diff
@@ -0,0 +1,414 @@
...
+#include <math.h>
+#include <unistd.h>
+
+#define bit(A,i) ((A >> i) & 1)
+#define min(a,b) ((a < b) ? a : b)
+#define max(a,b) ((a > b) ? a : b)
+#define bit_a_b(X,a,b) (((X << (sizeof(X)*8 - max(a,b) - 1)) >> (sizeof(X)*8 - max(a,b) - 1)) & ((X >> min(a,b) ) << min(a,b)))
+
+typedef unsigned long long int u_64;
+typedef unsigned int u_32;
```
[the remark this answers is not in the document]
Avaquet: Yes it's true

tests/test_cancellation/makefile
```diff
@@ -0,0 +1,5 @@
+test_canc_detect: test_canc_detect.c
+	gcc test_canc_detect.c -o test_canc_detect -Wall -O3 -ftree-vectorize
```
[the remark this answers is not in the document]
Avaquet: No he is not

src/backends/interflop-cancellation/interflop_cancellation.c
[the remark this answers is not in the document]
Avaquet: This part will be completely changed soon, we are gonna follow the advice from @pablooliveira.

src/backends/interflop-cancellation/interflop_cancellation.c
```diff
@@ -0,0 +1,280 @@
...
+static float _mca_sbin(float a, float b, int qop);
+static double _mca_dbin(double a, double b, int qop);
+
+/******************** MCA CONTROL FUNCTIONS *******************
+ * The following functions are used to set virtual precision and
+ * MCA mode of operation.
+ ***************************************************************/
+
+static int _set_mca_tolerance(int tolerance) {
+  MCALIB_T = tolerance;
```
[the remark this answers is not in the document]
Avaquet: Look at "parse_opt" function and particulary to the line 217, maybe i'm wrong but it's what we do ?

src/backends/interflop-cancellation/interflop_cancellation.c
Avaquet: When b is negative and a is positive we agree ,but when a is negative and b positive, it still an addition , isn't it ?

pablooliveira: In both cases a cancellation can happen.

src/backends/interflop-cancellation/interflop_cancellation.c
pablooliveira: Actually after thinking about this, the more realistic noise is the one you have used (MCA noise).
Indeed cancellation can produce negative and positive errors, when doing a - b. If a is exact and b is imprecise. After cancellation we will have a negative term.

Therefore, we won't do the optimization I was suggesting above. Your first version is fine !

src/backends/interflop-cancellation/interflop_cancellation.c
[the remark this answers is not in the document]
pablooliveira: Agreed

tests/test_cancellation/makefile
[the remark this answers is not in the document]
pablooliveira: Removed.

tests/test_ddebug_canc/archimedes.c
pablooliveira: Indeed.

src/backends/interflop-cancellation/interflop_cancellation.c
pablooliveira: I have changed your macros to use the new macros added by @yohanchatelain in #152. A small diffrence is that while mca backend uses a float128 to accumulate noise and rounds at the end. Here we do it directly in float64. But I think in the context of cancellation this is ok; since there is only one noise term. @yohanchatelain what do you think ?

pablooliveira:
@KillianBabilotte, @Avaquet, @DamDem, @redfoxp could you please review my changes and fix the change in README requested by @yohanchatelain ?

@yohanchatelain, if you have time, could you please review my changes. In particular do you think it is acceptable to add noise in the cancelled bits without using a quad. My reasoning is the following since the end of X is zeroes anyway then it should not change much to downsample to double before or after adding the noise; but we may have a small bias with the rounding. The advantage is that we can do the operation in double directly. What do you think ?

README.md
```diff
@@ -419,6 +420,29 @@ The option `--seed` fixes the random generator seed. It should not
...
+
+Finally the user should know that this backend is still experimental and in
+developpement. There is no guarantee on the results provided by this backend.
```
pablooliveira: Remove the "no guarantee clause"; the first sentence should be enough caveat.

README.md
```diff
@@ -419,6 +420,29 @@ The option `--seed` fixes the random generator seed. It should not
 generally be used except to reproduce a particular Bitmask
 trace.
 
+### Cancellation Backend (libinterflop_cancellation.so)
+
+The Cancellation backend implements an automatic cancellation detector at
+runtime. It is founded on bitwise operations to detect cancellation faster than
```
pablooliveira: This is incorrect. bitwise operations are not used to detect cancellation; we use difference in exponents.

KillianBabilotte: My bad. I was pretty sure that we use some operations on the bit representation of floating numbers to get their exponents and that I can call that bitwise operation but in any case I made a bad translation. But now I have a trouble to understand the code to get the exponent... 
`inline int32_t _get_exponent_binary64(const double d) {
  binary64 x = {.f64 = d};
  /* Substracts the bias */
  return x.ieee.exponent - DOUBLE_EXP_COMP;
}`
In this code or respectively the function for float case and 128bits float I don't understand the `binary64 x = {.f64 = d};` part. Did you have some references with whom I can understand what is happening here ?

src/backends/interflop-cancellation/interflop_cancellation.c
KillianBabilotte: I personally don't get how work the macro like I mention [here](https://github.com/verificarlo/verificarlo/pull/153#discussion_r389141728) but I agree with you on the use of float64 instead of float128 in this context. @Avaquet ?

README.md
KillianBabilotte: Done. But I have seen that the format in the current output of option `--h` isn't homogeneous in other backends. Some options description have a capital letter to begin and some don't.

README.md
pablooliveira: You did use bitwise operations actually; but I changed that to use #152 macros.

* binary64 is a union between a double and a struct with a set of bitfields corresponding to the sign, exponent, and mantissa.

* ```binary64 x = {.f64 = d};``` this copies the double to the union.
* ```x.ieee.exponent``` this extracts the exponent from the union. It is probably done with bit operations at the assembly level; but most backends do that already.

PR #170, opened 2020-07-08
Move tests from travis to docker compose

pablooliveira:
Travis is not running some of the PR tests; marking the requests as "abuse".

This seems a good time to merge building the docker image and doing the tests. It will also allow us to move to github-actions for our CI infrastructure which seems better integrated.


pablooliveira:
@Avaquet merging this should fix the test check in your PR.

Dockerfile
```diff
@@ -41,9 +41,8 @@ RUN ./autogen.sh && \
     		--with-flang CC=gcc-${GCC_VERSION} CXX=g++-${GCC_VERSION} \
     || cat config.log
 
-# Build and test verificarlo
-RUN make && make install && \
-    make installcheck
+# Build verificarlo
+RUN make && make install
```
titeup, line 45: I don't get why you removed installcheck here.

pablooliveira:
Thanks for the review.

> They are some (not impacting) error message in the github action log such as: https://github.com/verificarlo/verificarlo/runs/851728978?check_suite_focus=true#step:3:1596

I have checked those; they are not a problem since the test script files are created correctly. But they are indeed a configure.ac problem (they are not linked to this PR, they were happening before I believe). Maybe we should open a separate minor issue in the tracker ?

> BTW Did you check if a verificarlo test fail that the github action script return the failure accordingly?

Yes but I checked it locally on my machine. A github test would be good.

> I didn't know these github actions I had to read documentation and examples, looks great, the log are easy to navigate and point to. A definite improvement over travis!

Indeed.

pablooliveira:
Btw I removed the make installcheck from the Dockerfile, to decouple build and test of the image. The test is now managed throught the docker-compose yml file. According to docker documentation, this is the best practice.

PR #172, opened 2020-08-03
Function instrumentation

Avaquet:
New version of the function instrumentation, fix the bug with intrinsics.

### DIFF
--- a/src/libvfcfuncinstrument/libVFCFuncInstrument.cpp
+++ b/src/libvfcfuncinstrument/libVFCFuncInstrument.cpp
@@ -46,6 +46,7 @@
 #include <iostream>
 #include <set>
 #include <stdio.h>
+#include <string>
 #include <utility>
 #include <vector>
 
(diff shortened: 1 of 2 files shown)


src/libvfcfuncinstrument/libVFCFuncInstrument.cpp
```diff
@@ -207,9 +208,13 @@ void InstrumentFunction(std::vector<Value *> MetaData,
     int i = 0;
     for (auto &args : FunctionArgs)
       hook->setArgOperand(i++, args);
+    hook->setCalledFunction(HookedFunction);
     ret = Builder.Insert(hook);
   } else {
-    ret = Builder.CreateCall(HookedFunction, FunctionArgs);
+    CallInst *call = CallInst::Create(HookedFunction, FunctionArgs);
+    call->setAttributes(HookedFunction->getAttributes());
```
titeup, line 215: You are cloning only the function attributes? From our discussion I understood we need to clone arguments' attributes. Or both. Mabe we can clarify during our next meeting.

titeup: Hum... thinking twice, since you are using the parameter list of the cloned function you have the same attributes... This copy is only necessary when you duplicate your parameters. and I cannot see it in your previous 'step'

titeup: We check with allan today in IR, the copy in step2 is indeed copying the attribute of the arguments. Case closed. the rest is good for me. but I prefer to have a second review before merging.

yohanchatelain: LGTM

yohanchatelain (approved):

PR #173, opened 2020-08-17
Absolute error

mistoan:
This pull-request adds a new option to the interflop-mca and interflop-vprec back-ends: specify the error target as an absolute value.
The option is controlled through two parameters: error-mode and max-abs-error-exponent.
The former allows switching between relative and absolute errors, or both.
The latter controls the error threshold, as the exponent to a power of 2.


src/backends/interflop-mca/interflop_mca.c
```diff
@@ -106,6 +113,16 @@ typedef enum {
 
 static const char *MCA_MODE_STR[] = {"ieee", "mca", "pb", "rr"};
 
+/* define the available error modes */
+typedef enum {
+  mca_err_mode_rel,
+  mca_err_mode_abs,
+  mca_err_mode_all,
+  _mca_err_mode_end_
```
pablooliveira, line 121: Why is this canary (`_mca_err_mode_end`) needed ?

src/backends/interflop-mca/interflop_mca.c
```diff
@@ -196,11 +213,11 @@ static __float128 _noise_binary128(const int exp) {
...
-  (MCALIB_MODE == mcamode_ieee) ||					\
-  /* Check that we are not in a special case */				\
-  (FPCLASSIFY(X) != FP_NORMAL && FPCLASSIFY(X) != FP_SUBNORMAL) ||	\
-  /* In RR if the number is representable in current virtual precision, */ \
-  /* do not add any noise if */						\
+  (MCALIB_MODE == mcamode_ieee) ||					                                   \
+  /* Check that we are not in a special case */				                         \
+  (FPCLASSIFY(X) != FP_NORMAL && FPCLASSIFY(X) != FP_SUBNORMAL) ||	           \
+  /* In RR if the number is representable in current virtual precision, */     \
+  /* do not add any noise if */						                                     \
```
pablooliveira, line 220: Please avoid white-space only changes.

src/backends/interflop-mca/interflop_mca.c
```diff
@@ -209,34 +226,44 @@ static __float128 _noise_binary128(const int exp) {
...
+      if (((t_context *)CTX)->relErr) {                                        \
+        const int32_t e_n_rel = e_a - (VIRTUAL_PRECISION - 1);                 \
+        const typeof(*X) noise_rel = _NOISE(*X, e_n_rel);                      \
+        *X = *X + noise_rel;                                                   \
+      }                                                                        \
+      if (((t_context *)CTX)->absErr) {                                        \
+        const int32_t e_n_abs = ((t_context *)CTX)->absErr_exp;                \
+        const typeof(*X) noise_abs = _NOISE(*X, e_n_abs);                      \
+        *X = *X + noise_abs;                                                   \
+      }                                                                        \
```
pablooliveira, line 247: Assuming we drop the combined mode. This can be factored.
The `if` can only be around e_n computation. Then the next two lines are the same in the two modes (relative / absolute).

src/backends/interflop-mca/interflop_mca.c
```diff
@@ -408,6 +439,32 @@ error_t parse_opt(int key, char *arg, struct argp_state *state) {
...
     break;
+  case KEY_ERR_MODE:
+    /* mca error mode */
+    if (strcasecmp(MCA_ERR_MODE_STR[mca_err_mode_rel], arg) == 0) {
+      ctx->relErr = true;
+      ctx->absErr = false;
+    } else if (strcasecmp(MCA_ERR_MODE_STR[mca_err_mode_abs], arg) == 0) {
+      ctx->relErr = false;
+      ctx->absErr = true;
+    } else if (strcasecmp(MCA_ERR_MODE_STR[mca_err_mode_all], arg) == 0) {
```
pablooliveira, line 450: I'm not sure when I would use the "all" mode. Maybe we can simplify this and include only either absolute or relative but never both.

src/backends/interflop-mca/interflop_mca.c
```diff
@@ -209,34 +226,44 @@ static __float128 _noise_binary128(const int exp) {
 
 /* Macro function that adds mca noise to X
    according to the virtual_precision VIRTUAL_PRECISION */
-#define _INEXACT(X, VIRTUAL_PRECISION)                                         \
+#define _INEXACT(X, VIRTUAL_PRECISION, CTX)                                    \
   {                                                                            \
     if (_MUST_NOT_BE_NOISED(*X, VIRTUAL_PRECISION)) {                          \
       return;                                                                  \
     } else {                                                                   \
       const int32_t e_a = GET_EXP_FLT(*X);                                     \
```
pablooliveira, line 234: Since e_a is not used in the absolute mode, it could be hoisted inside the following `if`.

src/backends/interflop-vprec/interflop_vprec.c
```diff
@@ -273,47 +290,97 @@ void _set_vprec_inst_mode(vprec_inst_mode mode) {
...
 
-  /* round to zero or set to infinity if underflow or overflow compare to
+  /* round to zero or set to infinity if underflow or overflow compared to
    * VPRECLIB_BINARY32_RANGE */
   int emax = (1 << (binary32_range - 1)) - 1;
   int emin = (emax > 1) ? 1 - emax : -1;
 
+  /* in absolute error mode, the error threshold also gives the possible
+   * underflow limit */
+  if ((currentContext->relErr == true) && (currentContext->absErr == true)) {
```
pablooliveira, line 307: As in MCA, I'm not use of when to use rel and abs modes at the same time. If we drop the combined mode, this can be simplified.

src/backends/interflop-vprec/interflop_vprec.c
```diff
@@ -273,47 +290,97 @@ void _set_vprec_inst_mode(vprec_inst_mode mode) {
...
+        /* vprec error mode all */
+        if ((-1) * currentContext->absErr_exp < binary32_precision)
+          a = handle_binary32_denormal(a, emin, aexp.u32,
+                                       (-1) * currentContext->absErr_exp);
+        else
+          a = handle_binary32_denormal(a, emin, aexp.u32, binary32_precision);
+      } else if (currentContext->absErr == true) {
+        /* vprec error mode abs */
+        a = handle_binary32_denormal(a, emin, aexp.u32,
+                                     (-1) * currentContext->absErr_exp);
```
pablooliveira, line 342: The -1 here is strange and should be explained.

pablooliveira (changes requested):
Two general comments:

- I do not feel that allowing relative and absolute mode at the same time makes much sense. One of the two errors is going to dominate and absorb the other. Removing that could make the code much more simpler in both MCA and VPREC backend. That would be my preference.

- The implementation seems overly complex in the vprec case. I wonder if we could compute first, an equivalent precision that corresponds to the given absolute error threshold. Once we have computed this precision we could reuse the code as before without any more changes ?

titeup:
> Two general comments:
> 
> * I do not feel that allowing relative and absolute mode at the same time makes much sense. One of the two errors is going to dominate and absorb the other. Removing that could make the code much more simpler in both MCA and VPREC backend. That would be my preference.
> * The implementation seems overly complex in the vprec case. I wonder if we could compute first, an equivalent precision that corresponds to the given absolute error threshold. Once we have computed this precision we could reuse the code as before without any more changes ?

We discuss the case of having both kind of error at the same time and reach the conclusion it can be usefull and not so complicated to implement. The idea is that relative error is constraining the format length and absolut value emulating fixed point loss of precision. with both you can come very close to simulating for instance 32bits fixed point from double precision code. But we can discuss F2F it during our next meeting.

For the code complexity I agree, what should happen is that you compute the place for rounding or introducing noise and then use the usual code with a local value for the noise/rounding position. There is few mask/values to rebuild each time, or tabulate if we realize it is to costly to rebuild them everytime. Having both abs and relative option is just about compute a min() between two integers.

ddavid42:
> > Two general comments:
> > 
> > * I do not feel that allowing relative and absolute mode at the same time makes much sense. One of the two errors is going to dominate and absorb the other. Removing that could make the code much more simpler in both MCA and VPREC backend. That would be my preference.
> > * The implementation seems overly complex in the vprec case. I wonder if we could compute first, an equivalent precision that corresponds to the given absolute error threshold. Once we have computed this precision we could reuse the code as before without any more changes ?
> 
> We discuss the case of having both kind of error at the same time and reach the conclusion it can be useful and not so complicated to implement. The idea is that relative error is constraining the format length and absolute value emulating fixed point loss of precision. with both you can come very close to simulating for instance 32bits fixed point from double precision code. But we can discuss F2F it during our next meeting.
> 
> For the code complexity I agree, what should happen is that you compute the place for rounding or introducing noise and then use the usual code with a local value for the noise/rounding position. There is few mask/values to rebuild each time, or tabulate if we realize it is to costly to rebuild them every time. Having both abs and relative option is just about compute a min() between two integers.

I believe that computing "absolute error" AFTER "relative error" can add information but I don't see how having both at the same time can be done and/or what added value it can bring. Therefore, if it makes code clearer, I'm joining Pablo's view on this point.

src/backends/interflop-mca/interflop_mca.c
mistoan: It's not actually needed.
But it made sense to have it, in order to have the same coding style as in the rest of the file.

src/backends/interflop-mca/interflop_mca.c
mistoan: Ok.

src/backends/interflop-mca/interflop_mca.c
mistoan: Ok.
Solved.

src/backends/interflop-vprec/interflop_vprec.c
mistoan: Explanation: on the command line, the magnitude of the absolute error is specified as a power of 2.
In this case, I made the unreasonable assumption that the power of two would be negative, as we're exploring precision...
But handle_binary32_denormal has as fourth parameter the precision of the new format, which is a positive number.

Should be changed to: abs(currentContext->absErr_exp)

src/backends/interflop-mca/interflop_mca.c
mistoan: This must have happened when clang-format complained...

src/backends/interflop-vprec/interflop_vprec.c
mistoan: I've made the changes.

src/backends/interflop-mca/interflop_mca.c
mistoan: I've removed it.
From interflop_vprec, as well.

mistoan:
There has been a refactoring of the code for the interflop_vprec back-end, which now regroups the computation of the virtual precision at which to round the inputs to the operations.

mistoan:
The proposed changes now have entries in the documentation of the corresponding back-ends.

pablooliveira:
@mistoan thanks for the changes. I'm reviewing the PR and I do not understand what is going on. You seem to have changes here that have already been merged. For example, when I look at the changes, I see changes for the release 4.0.1 such as in the CHANGELOG. 

I think something strange has happened during your rebase. But in the current form, it is hard to see what are the *new* changes here. Could you please fix the PR please so it only shows your changes ?

pablooliveira:
Ok I have updated the base branch so github only shows commits with current master.

src/backends/interflop-vprec/interflop_vprec.c
```diff
@@ -273,34 +378,87 @@ void _set_vprec_inst_mode(vprec_inst_mode mode) {
...
 
-  /* round to zero or set to infinity if underflow or overflow compare to
+  /* round to zero or set to infinity if underflow or overflow compared to
    * VPRECLIB_BINARY32_RANGE */
   int emax = (1 << (binary32_range - 1)) - 1;
-  // here emin is the smallest exponent in the *normal* range
+  /* here emin is the smallest exponent in the *normal* range */
   int emin = 1 - emax;
 
+  // /* in absolute error mode, the error threshold also gives the possible
```
pablooliveira, line 394: Why do we keep this commented block ?

src/backends/interflop-vprec/interflop_vprec.c
```diff
@@ -273,34 +378,87 @@ void _set_vprec_inst_mode(vprec_inst_mode mode) {
...
+      } else {
+        /* relative error mode */
+        a = handle_binary32_denormal(a, emin, binary32_precision);
+      }
     }
   } else {
-    a = round_binary32_normal(a, binary32_precision);
+    /* else, normal case: can be executed even if a
+     previously rounded and truncated as denormal */
+    if (currentContext->absErr == true) {
```
pablooliveira, line 437: This block and the counterpart for 64 below should be factored in one or two separate functions.

src/backends/interflop-vprec/interflop_vprec.c
```diff
@@ -242,6 +258,95 @@ void _set_vprec_inst_mode(vprec_inst_mode mode) {
   }
 }
 
+/******************** VPREC HELPER FUNCTIONS *******************
+ * The following functions are used to set virtual precision,
+ * VPREC mode of operation and instrumentation mode.
+ ***************************************************************/
+
+int compute_absErr_vprec_binary32(bool isDenormal, void *context, int expDiff,
```
pablooliveira, line 266: These two functions, can be made one, by passing the mantissa size as an argument. It is the only difference between them both.

src/backends/interflop-vprec/interflop_vprec.c
```diff
@@ -939,6 +1177,9 @@ static error_t parse_opt(int key, char *arg, struct argp_state *state) {
 static struct argp argp = {options, parse_opt, "", "", NULL, NULL, NULL};
 
 void init_context(t_context *ctx) {
+  ctx->relErr = true;
+  ctx->absErr = false;
+  ctx->absErr_exp = 112;
```
pablooliveira, line 1182: Why 112 ? We need a comment here or a CONSTANT.

pablooliveira (changes requested):
Thanks:
- Documentation is clear and well
- I have not yet looked at tests carefully
- for mca the code added is quite straighforward.
- for vprec, there is too much code added and the changes make vprec much harder to read and maintain. I still think supporting the "all" mode is making things overly complex for little added benefit. I have included some suggestion to factor code.

src/backends/interflop-vprec/interflop_vprec.c
mistoan: I've replaced this initialization with the corresponding constant from float_const.h

src/backends/interflop-vprec/interflop_vprec.c
mistoan: There's no need to keep this block.
It was an oversight.
Now removed, from both _vprec_round_binary32 and _vprec_round_binary64.

pablooliveira:
@mistoan please tell me when this is done so I can do a last review and merge !

src/backends/interflop-vprec/interflop_vprec.c
mistoan: This block has been factorized into the handle_binary32_normal_absErr and handle_binary64_normal_absErr functions.

src/backends/interflop-vprec/interflop_vprec.c
mistoan: I think that regrouping the two functions into a single one would obfuscate the code too much and render code maintainance more difficult.
I have identified a case that was never used and removed it, so this has simplified a bit the code.

mistoan:
@pablooliveira I have made the required changes.
I have also rebased the branch to include all recent updates in the main branch.

mistoan:
@pablooliveira I've rebased the code to integrate the latest changes in the main branch, so everything should be up to date now.

pablooliveira:
Closing this. Continues in #198.

PR #174, opened 2020-08-20
Searching and fixing vprec bugs

titeup:
Hi all, 
I started this pull request to share progress in search of our vprec bug.
The vprec backend benchmark had a bug preventing to test negative exponent value and the log was not giving all info to reproduce found error. This part is fixed with the first commit I made before starting the pull request.
The remaining problem in the bench is that it is making a poor usage of numpy.random.normal, it is not covering many cases of the code, it is slow, and finally despite non empty error log, the make installcheck command receive a positive feedback.
All of this will be fixed soon.

And then comes the vprec backend issue to investigate which can be illustrated with the following log (mpfr has the correct answer):
float: MODE=OB RANGE=2 PRECISION=3 OP=+
Relative error too high: a=0x1.ff10ddbd996a2p-7 b=0x1.137d3527080f0p-2
 mpfr=0.25 vprec=0.3125 error=0.25 (2.0 b=2)

I will open an issue for this one.


pablooliveira:
Excellent, that you found a reduced test case. Congrats !

ddavid42:
One comment: 

>And then comes the vprec backend issue to investigate which can be illustrated with the following log (mpfr has the correct >answer):
>float: MODE=OB RANGE=2 PRECISION=3 OP=+
>Relative error too high: a=0x1.ff10ddbd996a2p-7 b=0x1.137d3527080f0p-2
>mpfr=0.25 vprec=0.3125 error=0.25 (2.0 b=2)

This case corresponds to rounding correctly the value 'b' (which is 0.25) as 'a' is way to small to be represented with the given parameters and have an impact on the addition.

titeup:
> One comment:
> 
> > And then comes the vprec backend issue to investigate which can be illustrated with the following log (mpfr has the correct >answer):
> > float: MODE=OB RANGE=2 PRECISION=3 OP=+
> > Relative error too high: a=0x1.ff10ddbd996a2p-7 b=0x1.137d3527080f0p-2
> > mpfr=0.25 vprec=0.3125 error=0.25 (2.0 b=2)
> 
> This case corresponds to rounding correctly the value 'b' (which is 0.25) as 'a' is way to small to be represented with the given parameters and have an impact on the addition.

This what I mean with mpfr has the correct answer ;)
Furthermore vprec results is too high by a forth of the value (0.0625), which means that we add a one in the pseudo mantissa at the last position for the given precision (3) which is pointing to a rounding calculation error. not exponent computation. But then I wonder why it is appearing only on results with negative exponents... anyway, there is a bug...

titeup:
> > One comment:
> > > And then comes the vprec backend issue to investigate which can be illustrated with the following log (mpfr has the correct >answer):
> > > float: MODE=OB RANGE=2 PRECISION=3 OP=+
> > > Relative error too high: a=0x1.ff10ddbd996a2p-7 b=0x1.137d3527080f0p-2
> > > mpfr=0.25 vprec=0.3125 error=0.25 (2.0 b=2)
> > 
> > 
> > This case corresponds to rounding correctly the value 'b' (which is 0.25) as 'a' is way to small to be represented with the given parameters and have an impact on the addition.
> 
> This what I mean with mpfr has the correct answer ;)
> Furthermore vprec results is too high by a forth of the value (0.0625), which means that we add a one in the pseudo mantissa at the last position for the given precision (3) which is pointing to a rounding calculation error. not exponent computation. But then I wonder why it is appearing only on results with negative exponents... anyway, there is a bug...
> EDIT: maybe I did a mistake in the following but: actually b=1.1 means 1.00010011 in binary adding 1/2 ulp with p=3 (bits in the mantissa) rounds to 1.00100011 which is 0.3125 if I am correct. maybe vprec is right finally... Why does mpfr round to 0.25 in that case? maybe we are not counting the implicit one and they do, and this why we also have a lot of error around 10^-7 and 10^-8 when p=23. I will check.

For MPFR precision=53 is equivalent to double => it means 52+1 implicit bit. And now I'll check in our code.

ddavid42:
> > One comment:
> > > And then comes the vprec backend issue to investigate which can be illustrated with the following log (mpfr has the correct >answer):
> > > float: MODE=OB RANGE=2 PRECISION=3 OP=+
> > > Relative error too high: a=0x1.ff10ddbd996a2p-7 b=0x1.137d3527080f0p-2
> > > mpfr=0.25 vprec=0.3125 error=0.25 (2.0 b=2)
> > 
> > 
> > This case corresponds to rounding correctly the value 'b' (which is 0.25) as 'a' is way to small to be represented with the given parameters and have an impact on the addition.
> 
> This what I mean with mpfr has the correct answer ;)
> Furthermore vprec results is too high by a forth of the value (0.0625), which means that we add a one in the pseudo mantissa at the last position for the given precision (3) which is pointing to a rounding calculation error. not exponent computation. But then I wonder why it is appearing only on results with negative exponents... anyway, there is a bug...
> EDIT: maybe I did a mistake in the following but: actually b=1.1 means 1.00010011 in binary adding 1/2 ulp with p=3 (bits in the mantissa) rounds to 1.00100011 which is 0.3125 if I am correct. maybe vprec is right finally... Why does mpfr round to 0.25 in that case? maybe we are not counting the implicit one and they do, and this why we also have a lot of error around 10^-7 and 10^-8 when p=23. I will check.

Hopp... be carefull because in that case (with range=2) it looks like you are in the equivalent of the "subnormal" case.
and 1.001 is not 0.3125 wich is 0.25 + 0.0625 (1.01p-2).

titeup:
> > EDIT: maybe I did a mistake in the following but: actually b=1.1 means 1.00010011 in binary adding 1/2 ulp with p=3 (bits in the mantissa) rounds to 1.00100011 which is 0.3125 if I am correct. maybe vprec is right finally... Why does mpfr round to 0.25 in that case? maybe we are not counting the implicit one and they do, and this why we also have a lot of error around 10^-7 and 10^-8 when p=23. I will check.
> 
> For MPFR precision=53 is equivalent to double => it means 52+1 implicit bit. And now I'll check in our code.


define VPREC_PRECISION_BINARY32_MAX FLOAT_PMAN_SIZE
....
if (VPREC_PRECISION_BINARY32_MAX < precision) {
    logger_error("invalid precision provided, "
....

=> Boum! now we have to decide what to do. My humble opinion is to take the same convention as mpfr and use mantissa size and not pseudo mantissa.

titeup:
> Hopp... be carefull because in that case (with range=2) it looks like you are in the equivalent of the "subnormal" case.
> and 1.001 is not 0.3125 wich is 0.25 + 0.0625 (1.01p-2).

So maybe there is an additional bug with shift computation...

pablooliveira:
> => Boum! now we have to decide what to do. My humble opinion is to take the same convention as mpfr and use mantissa size and not pseudo mantissa.
> 
> EDIT: seems like it has been taken into account in the benchmark: for mpfr computation: prec = get_var_env_int(PRECISION)**+1**

Good. I see your EDIT, so this is probably not the issue anymore.

Just in case, changing the convention between pseudo-mantissa / mantissa, would require also changing it in MCA (because we should be coherent across our backends, so that precision always refers to the same bit). And I would prefer not to change in MCA to keep consistent with Parker's and our published works.

titeup:
> y = 0x1.b23058p-3 = 0x1. 1011 0010 0011 0000 1001 100 p-3 = 0x0. 0011 0110 0100 0110 0001 0011 00 p-1
> up(y)   = 0x0. 0011 0110 0100 0110 0001 010 p-1 = 0x1. 1011 0010 0011 0000 1010 p-3 = 0x1.d2306 p-3 = mpfr
> down(y) = 0x0. 0011 0110 0100 0110 0001 001 p-1 = 0x1. 1011 0010 0011 0000 1001 p-3 = 0x1.d2305 p-3 = vprec

Seems like we found the bug for that particular one. Thanks David for pointing to denormals ;) PR update to come soon.
I'll work on the shorter precision one now.

titeup:
> > y = 0x1.b23058p-3 = 0x1. 1011 0010 0011 0000 1001 100 p-3 = 0x0. 0011 0110 0100 0110 0001 0011 00 p-1
> > up(y)   = 0x0. 0011 0110 0100 0110 0001 010 p-1 = 0x1. 1011 0010 0011 0000 1010 p-3 = 0x1.d2306 p-3 = mpfr
> > down(y) = 0x0. 0011 0110 0100 0110 0001 001 p-1 = 0x1. 1011 0010 0011 0000 1001 p-3 = 0x1.d2305 p-3 = vprec
> 
> Seems like we found the bug for that particular one. Thanks David for pointing to denormals ;) PR update to come soon.

There was a bug, but It was not enough fix everything. WIP...

yohanchatelain:
**Update**
I think that there is a problem of double rounding.
MPFR does not have the same rounded value if the value comes from a computation or if it is directly rounded.
In the following example, we can see that first [1] MPFR returns **-0x1.ef1eb40000000p-1** which is the result of the operation rounded to the desired precision. Now, the result of this operation in binary32 is **+0x1.ef1eb6p-1**. If we ask [2] MPFR to round the value for the given format, we have **0x1.ef1eb80000000p-1** which is what VPREC also gives.
And if we do the same [3] computation in binary64, we have the same result as MPFR: **-0x1.ef1eb40000000p-1**.

[1]
```
VERIFICARLO_VPREC_MODE=OB VERIFICARLO_PRECISION=23 VERIFICARLO_VPREC_RANGE=2 VFC_BACKENDS="libinterflop_vprec.so --mode=ob --precision-binary32=${VERIFICARLO_PRECISION} --range-binary32=${VERIFICARLO_VPREC_RANGE} --precision-binary64=${VERIFICARLO_PRECISION} --range-binary64=${VERIFICARLO_VPREC_RANGE}" ./comparison.sh  -0x1.6cb1aep+0 0x1.5b8db4p-1 x float
VFC_BACKENDS=libinterflop_vprec.so --mode=ob --precision-binary32=23 --range-binary32=2 --precision-binary64=23 --range-binary64=2
[MPFR] (before rounding) a=-0x1.6cb1aep+0
[MPFR] (before rounding) b=+0x1.5b8db4p-1
MPFR context: Context(precision=24, emax=2, emin=-22, subnormalize=True, rounding=ROUND_TIES_TO_EVEN)
MPFR mul -0x1.6cb1ae0000000p+0 0x1.5b8db40000000p-1 = -0x1.ef1eb40000000p-1
MPFR mul -0x1.6cb1ae0000000p+0 0x1.5b8db40000000p-1 = -0x1.ef1eb40000000p-1
[MPFR]  a=-0x1.6cb1aep+0
[MPFR]  b=+0x1.5b8db4p-1
[MPFR]  res=-0x1.ef1eb4p-1
-0x1.ef1eb40000000p-1
loaded backend libinterflop_vprec.so
load backend with precision-binary32 = 23, range-binary32 = 2, precision-binary64 = 23, range-binary64 = 2, mode = ob, daz = false, ftz = false and instrument = none
[VPREC] input: -0x1.ef1eb6p-1 
[VPREC] res: -0x1.ef1eb8p-1
```

[2]
```
VERIFICARLO_VPREC_MODE=OB VERIFICARLO_PRECISION=23 VERIFICARLO_VPREC_RANGE=2 VFC_BACKENDS="libinterflop_vprec.so --mode=ob --precision-binary32=${VERIFICARLO_PRECISION} --range-binary32=${VERIFICARLO_VPREC_RANGE} --precision-binary64=${VERIFICARLO_PRECISION} --range-binary64=${VERIFICARLO_VPREC_RANGE}" ./comparison.sh  0x0 0x1.ef1eb6p-1 + float
VFC_BACKENDS=libinterflop_vprec.so --mode=ob --precision-binary32=23 --range-binary32=2 --precision-binary64=23 --range-binary64=2
[MPFR] (before rounding) a=0
[MPFR] (before rounding) b=-0x1.ef1eb6p-1
[MPFR]  a=0
[MPFR]  b=-0x1.ef1eb6p-1
[MPFR]  res=-0x1.ef1eb8p-1
-0x1.ef1eb80000000p-1
loaded backend libinterflop_vprec.so
load backend with precision-binary32 = 23, range-binary32 = 2, precision-binary64 = 23, range-binary64 = 2, mode = ob, daz = false, ftz = false and instrument = none
[VPREC] input: -0x1.ef1eb6p-1 
[VPREC] res: -0x1.ef1eb8p-1
```

[3]
```
VERIFICARLO_VPREC_MODE=OB VERIFICARLO_PRECISION=23 VERIFICARLO_VPREC_RANGE=2 VFC_BACKENDS="libinterflop_vprec.so --mode=ob --precision-binary32=${VERIFICARLO_PRECISION} --range-binary32=${VERIFICARLO_VPREC_RANGE} --precision-binary64=${VERIFICARLO_PRECISION} --range-binary64=${VERIFICARLO_VPREC_RANGE}" ./comparison.sh  -0x1.6cb1aep+0 0x1.5b8db4p-1 x double
VFC_BACKENDS=libinterflop_vprec.so --mode=ob --precision-binary32=23 --range-binary32=2 --precision-binary64=23 --range-binary64=2
[MPFR] (before rounding) a=-0x1.6cb1ae00000000p+0
[MPFR] (before rounding) b=+0x1.5b8db400000000p-1
MPFR context: Context(precision=24, emax=2, emin=-22, subnormalize=True, rounding=ROUND_TIES_TO_EVEN)
MPFR mul -0x1.6cb1ae0000000p+0 0x1.5b8db40000000p-1 = -0x1.ef1eb40000000p-1
MPFR mul -0x1.6cb1ae0000000p+0 0x1.5b8db40000000p-1 = -0x1.ef1eb40000000p-1
[MPFR]  a=-0x1.6cb1ae00000000p+0
[MPFR]  b=+0x1.5b8db400000000p-1
[MPFR]  res=-0x1.ef1eb4p-1
-0x1.ef1eb40000000p-1
loaded backend libinterflop_vprec.so
load backend with precision-binary32 = 23, range-binary32 = 2, precision-binary64 = 23, range-binary64 = 2, mode = ob, daz = false, ftz = false and instrument = none
[VPREC] input: -0x1.ef1eb523c4580p-1
[VPREC] res: -0x1.ef1eb4p-1
```

titeup:
I will look into the double rounding issue.
We fixed other bugs in the meantime, but when runing larger test I realize that inbound and FULL modes are provoking a ot of new ones, still looks to be on denormals only.

pablooliveira:
As discussed during the meeting, our changes in 30da478, made obsolete the xexp argument in `handle_denormals` and `round_denormals`. Removing this argument will simplify the code.

titeup:
> As discussed during the meeting, our changes in [30da478](https://github.com/verificarlo/verificarlo/commit/30da4782217302286bb5cd5f500d151b94097bc8), made obsolete the xexp argument in `handle_denormals` and `round_denormals`. Removing this argument will simplify the code.

I started the process, we can suppress one level of function, and in interflop_vprec.c, it seems that we can improve thing the same way by using .ieee.exponent and complement instead of masks and shift.

titeup:
Now that we are testing much more combination of mode (IB/FULL is problematic), range and precision we have new set of faulty cases... The cases generation for double precision needs improvement too, see comment in the source code of generate input.py
Still some work...

pablooliveira:
Regarding 30da478, it was a useful step to get the correct shifts in the denormal case. Nevertheless, I have though about the new implementation yesterday; and I believe the previous method where half-ulp was encoded as the sum of two terms has advantages even in the case of denormals.

Indeed when xexp is a very small power (close to the representable limit of the containing format), the new method has a chance to absorb the computed half-ulp since the power is too small.

I have written on paper the correct implementation using a combination of two-terms, it is not difficult. I will implement it. We can disscus during the afternoon meeting.

titeup (commented):
Some cases generates in check_output are generating infinite relative error despite the logicl statement above should avoid it. I setup an error message and early exit but didn't fix the bug yet.
Generate input should be fine now.
We have some cases where mpfr round to 0 and vprec to inf => we probably have a bug in overflow/underflow management confusing between the two cases.

pablooliveira:
With latest changes, test_vprec_backend and test_vprec_backend_simple are passing.
To sum up, there was two real bugs remaining:
1. possible overflow in denormals exponent that was fixed by moving this case to float128 (performance is not so important on denormals)
2. shift-by-one problem in emin for denormals

We also realized that we do not follow the rounding ties to even convention; currently we only guarantee faithful rounding. Therefore, checks were changed to account for 1-ulp error in results.

There is still one remaining bug with 
     FAIL: testplan.test 31 ./test_vprec_instrument

@Avaquet if you want to have a look that would be great, since you know that test best. I suspect we just need to update the expected outputs.

tests/test_vprec_backend/generic_test.sh
```diff
@@ -61,59 +67,53 @@ print_sep() {
 }
 
 compute_op() {
-    rm mpfr.txt
-    rm vprec.txt
     while read a b; do
 	./compute_mpfr_rounding.py $a $b $1 $2 >> mpfr.txt
 	./compute_vprec_rounding $a $b $1 >> vprec.txt
-    done < input.txt
+	done < input.txt
```
titeup, line 73: I suppose this is still working, but I would say the previous indentation was the correct one.

tests/test_vprec_backend/generic_test.sh
```diff
@@ -31,7 +37,7 @@ float_type_list=("float" "double")
 
 # Modes list
 if [ $USECASE = "fast" ]; then
-    modes_list=("OB")
+    modes_list=("FULL")
```
titeup, line 40: As we discuss during last friday meeting, maybe we can set to OB by default for the fast test in order to save testing time on PR push. 
The rational is that OB alone is enough to validate the rounding functions.
However full is also testing the computation part of the backend, which is bugfree and not suppose to move, but who knows... 
In a nutshell I am ok to go back on "OB". Anyobjection from the other PR contributors?

titeup (commented):
Someone else will have to review and approve since I opened this pullrequest I am not allowed to give approval myself.
Thanks, Eric

tests/test_vprec_backend/generic_test.sh
titeup: done

tests/test_vprec_backend/generic_test.sh
titeup: done

pablooliveira (approved):
LGTM.

PR #178, opened 2020-08-27
Take last field from gcc version call to support other linux distros.

roblim1:


pablooliveira (approved):
This looks good to me. Thanks for the contribution @roblim1 !

PR #182, opened 2020-09-14
Add exclude mode to delta-debug

pablooliveira:
Adds an exclude mode to delta-debug. 

If the environment variable VFC_DDEBUG_EXCLUDE is defined and points to a file containing a list of addresses, those addresses will never be instrumented or generated during delta-debug generation run. 

This variable operates _both_ at generation and filtering stages.

An example and test is available in tests/test_ddebug_exclude.

@mistoan I think this is what you requested this morning to be able to optimize absolute error exploration. Could you please check that it is satisfactory ?


tests/test_ddebug_canc/test.sh
```diff
@@ -1,5 +1,6 @@
 #!/bin/bash
 
+make clean
```
mistoan, line 3: This commit shouldn't be in this pull request.
But it's needed addition to verificarlo.

mistoan (changes requested):
This looks great.

Glad to see that you managed to factor the code for the include and the exclude files.

There's a problem with one commit which isn't concerning this feature.

pablooliveira:
> There's a problem with one commit which isn't concerning this feature.

The minor change is a fix to tests, it does not affect the CI runner but is nicer when running local tests since it ensures that the program is recompiled.
It *is actually* related to this PR, since I discovered it when testing an intermediate faulty version of this PR and realized only some of the delta-debug tests were triggered.

It's ok to include minor fixes like this in larger PRs.

mistoan:
> > There's a problem with one commit which isn't concerning this feature.
> 
> The minor change is a fix to tests, it does not affect the CI runner but is nicer when running local tests since it ensures that the program is recompiled.
> It _is actually_ related to this PR, since I discovered it when testing an intermediate faulty version of this PR and realized only some of the delta-debug tests were triggered.
> 
> It's ok to include minor fixes like this in larger PRs.

Ok, this makes sense.
Then for what I'm concerned, I think it's ready for merging.

mistoan (approved):

pablooliveira:
Thanks for the review and merge.

PR #183, opened 2020-09-29
Fix error with .ll path during compilation

yohanchatelain:
- Previous version uses the absolute path of the source file to generate 
  intermediate files (.ll). But this path can be not accessible to the 
  user (read-only). We use the current directory instead.


pablooliveira (approved):
Looks good to me.

PR #185, opened 2020-10-19
Pre release 0.4.1

yohanchatelain:
- Add wildcard support in include/exclude file to match any pattern in a regex
  Now we can have requests like `dir/* function_*` to match any function starting by `f` of any modules in `dir`
- Increase the size of the supported vectors from 4 to 16


pablooliveira (approved):
Looks good to me.

I have some bullet points to add to CHANGELOG before doing the 0.4.1 release... I'll try to be quick.

PR #179, opened 2020-09-07
Precision exploration

Avaquet:
New features:
- More complete test for instrumentation and vprec
- Custom function set exploration with vfc_precexp
- Automatic report with vfc_report
- Pointer interception during function call instrumentation
- Fetch of value interval for each intercepted argument

PR #189, opened 2020-11-03
CallInst and MDNode scope safe

roblim1:
During Function pass, if the metadata does not exist for the CallInst, the pass continues, which crashes when linking shared libraries.


pablooliveira:
Thanks @roblim1. I tag @Avaquet on this PR which was the main author of FuncInstrument.

pablooliveira:
After going through your PR and discussing with @Avaquet, this looks quite good. Thanks for the catch and the contribution !

In the improvements to FuncInstrument in #179, @Avaquet does not uses the metadata in the same way, so this should not trigger. But until #179 is fixed and merged; we should merge your PR.

Currently the tests are failing because the changes are not formatted according to clang c++ style.
If you want to fix this yourself, you can just run clang-format -i on `FuncInstrument.cpp` and commit the changes.

Moreover 0d85c0a has been already merged into master. This needs to be rebased on current master.

If you want, I can do these last two points for you. Just tell me what you prefer.

roblim1:
> If you want, I can do these last two points for you. Just tell me what you prefer.

Yes, if you could rebase head and reformat according to clang style, I'd appreciate it.  Thanks @pablooliveira !

PR #199, opened 2021-01-17
Do not instrument unsupported types, such as function pointers.

roblim1:
Otherwise, vfc_enter_function gets invoked during program execution, and an entry not created will cause the program to crash.


roblim1:
Specifically, when doing the lookup here:  https://github.com/verificarlo/verificarlo/blob/cdd09a76f8865347d1d4a9964b3a6ecb51a05245/src/common/vfc_hashmap.c#L212

The map is uninitialized, since an entry was not created for that argument.  The lookup assumes that the entry exists and will succeed.

PR #201, opened 2021-02-07
clean tests folder after installcheck

Sholde:
- add a target for generate a list with git
- add a targe to clean the folder with the list
- provide a base list


Sholde:
I will open a new PR later.

PR #214, opened 2021-03-08
Fix test that requires --save-temps

pablooliveira:
- test_ieee_backend requires ```--save-temps``` since it inspects .ll files.


yohanchatelain (changes requested):
Thanks @pablooliveira you for finding the bug. You should also replace `test.2.ll` by `test*.2.ll` since the .ll name is random.

pablooliveira:
Thanks @yohanchatelain !

yohanchatelain (approved):
:+1:

PR #216, opened 2021-03-08
Remove demangle calls from LLVM passes (see #206).

### LINKED ISSUE #206: Inconsistency in the way instrumented functions are specified for inclusion/exclusion and --inst-func
There is an inconsistency between the way in which functions are specified for inclusion/exclusion from verificarlo instrumentation and for the --inst-func option.
For inclusion/exclusion the mangled name is used, but for --inst-func option the un-mangled name is used.

pablooliveira:
Closes #206.


mistoan (approved):
This seems to remove all references to demangling, indeed.
I've rechecked and these calls were the last ones remaining.

PR #226, opened 2021-03-18
Fix llvm version error

Sholde:
Hello,

When I tried to use Verificarlo with llvm version 11.1.0 I see this message:
```
LLVM version up to 11.1.0 is supported
```
I think it is just one forgets.

I also add **pandas** dependency on README.

**UPDATE:**
The message is not:
```
LLVM version up to 11.1.0 is supported
```
But it is:
```
LLVM version up to 11.0.1 is supported
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,7 +72,7 @@ Please ensure that Verificarlo's dependencies are installed on your system:
   * LLVM, clang and opt from 4.0 up to 11.0.1, http://clang.llvm.org/
   * gcc from 4.9
   * flang for Fortran support
-  * python3 with numpy and bigfloat packages
+  * python3 with numpy, bigfloat and pandas packages
   * autotools (automake, autoconf)
 
 Then run the following command inside verificarlo directory:
@@ -92,7 +92,7 @@ install procedure:
 ```bash
    $ sudo apt-get install libmpfr-dev clang-7 flang-7 llvm-7-dev \
        gcc-7 autoconf automake build-essential python3 python3-numpy python3-pip
-   $ sudo pip3 install bigfloat
+   $ sudo pip3 install bigfloat pandas
    $ cd verificarlo/
    $ ./autogen.sh
    $ ./configure --with-flang CC=gcc-7 CXX=g++-7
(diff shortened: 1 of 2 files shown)


pablooliveira:
Dear @Sholde, thanks for your contribution.
The pandas dependency looks good and can be merged.

But the fix with the LLVM version is not correct. The message is correctly worded, but maybe the version check has a problem.

pablooliveira (changes requested):
The message in ax_llvm is correct but maybe there is a problem elsewhere with the version checking.

Sholde:
Ok I will try to catch it.

pablooliveira:
After your update it seems everything is fine and we can close this PR. Could you please confirm @Sholde and close the PR if you agree ?

pablooliveira:
To fix the issue you just have to update the LLVM highest-supported version number in configure.ac. Instead of using 11.0.1 you can replace it with 11.9 for example but please check that it works and tests are successful before submitting a PR for this.

Sholde:
I just reinstall **llvm10** for work. I don't think reinstall **llvm 11.1.0** for the moment.

Else some tests don't pass in my machine even if **llvm10**. Maybe I haven't some dependencies.
But for your information all tests which passed with **llvm10**, passed with **llvm 11.1.0**.

But I wonder yet why the message specify that `LLVM version up to 11.0.1 is supported` although it is not supported 🤔.

I close this PR.

pablooliveira:
If I understand the message you tried with llvm 11.1.0 and got the message that only "llvm up to 11.0.1 is supported" which makes sense to me because 11.1.0 > 11.0.1

Sholde:
Yes but verificarlo support until llvm 11.0.1. So I should have "llvm up to 11.0.1 is not supported" I think. Because the configuration script crash at this time.

pablooliveira:
No, what the message is telling you is the maximum LLVM version that we support. So in my opinion the message is correct. Thanks for the report though !

Sholde:
Oh yes, I think I need to improve my english. Sorry ^^.

PR #228, opened 2021-03-29
fix pandas and libtool dependencies

Sholde:
Hello,
I reinstall Verificarlo on another sytem running Linux Mint, and after used the following install procedure in README, the configure script doesn't find `llvm-config` file.
I check in stackoverflow, and I find that we need `libtool` package.

I also add pandas dependency of #226 because I don't know where is the problem with llvm version.

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,7 +72,7 @@ Please ensure that Verificarlo's dependencies are installed on your system:
   * LLVM, clang and opt from 4.0 up to 11.0.1, http://clang.llvm.org/
   * gcc from 4.9
   * flang for Fortran support
-  * python3 with numpy and bigfloat packages
+  * python3 with numpy, bigfloat and pandas packages
   * autotools (automake, autoconf)
 
 Then run the following command inside verificarlo directory:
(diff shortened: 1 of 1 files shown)


yohanchatelain:
Do you have the log of your configure? Have you checked if llvm-config-<version> exists? Sometime the binary `llvm-config` is not created, only the one postfixed with the version.

yohanchatelain:
And why do we need `pandas`?

Sholde:
Yes I had a llvm-config in my path but the configuration script don't find it, I don't know why. I just know that installing `libtool` fix the problem. And I don't save my log.

For `pandas` I don't know, as I said I reinstall in other machine `verificarlo` and the configuration script said me that I need `pandas`.

titeup:
panda has been used by @Avaquet we can try to remove the depencie in future version but right now it is in use:

https://github.com/verificarlo/verificarlo/blame/eb4dab174b99690000d7383b04d1c55343afe225/src/vfc_report#L3

yohanchatelain:
> panda has been used by @Avaquet we can try to remove the depencie in future version but right now it is in use:
> 
> https://github.com/verificarlo/verificarlo/blame/eb4dab174b99690000d7383b04d1c55343afe225/src/vfc_report#L3

Ok that's fine, I didn't follow that

pablooliveira (approved):

pablooliveira:
Thanks !

PR #230, opened 2021-04-01
Added fix for vprec backend, where precision and range should be set …

roblim1:
…to the user specified option, and if none is passed in, the options are set to default.  Also, the precision and range of an argument should be used instead of the default values.

### DIFF
--- a/src/backends/interflop-vprec/interflop_vprec.c
+++ b/src/backends/interflop-vprec/interflop_vprec.c
@@ -777,10 +777,10 @@ void _interflop_enter_function(interflop_function_stack_t *stack, void *context,
     function_inst->isIntrinsicFunction = function_info->isIntrinsicFunction;
     function_inst->useFloat = function_info->useFloat;
     function_inst->useDouble = function_info->useDouble;
-    function_inst->OpsRange64 = VPREC_RANGE_BINARY64_DEFAULT;
-    function_inst->OpsPrec64 = VPREC_PRECISION_BINARY64_DEFAULT;
-    function_inst->OpsRange32 = VPREC_RANGE_BINARY32_DEFAULT;
-    function_inst->OpsPrec32 = VPREC_PRECISION_BINARY32_DEFAULT;
+    function_inst->OpsRange64 = VPRECLIB_BINARY64_RANGE;
+    function_inst->OpsPrec64 = VPRECLIB_BINARY64_PRECISION;
+    function_inst->OpsRange32 = VPRECLIB_BINARY32_RANGE;
+    function_inst->OpsPrec32 = VPRECLIB_BINARY32_PRECISION;
     function_inst->nb_input_args = 0;
     function_inst->input_args = NULL;
     function_inst->nb_output_args = 0;
@@ -843,13 +843,11 @@ void _interflop_enter_function(interflop_function_stack_t *stack, void *context,
       function_inst->input_args[i].min_range = INT_MAX;
       function_inst->input_args[i].max_range = INT_MIN;
       function_inst->input_args[i].exponent_length =
-          (type == FDOUBLE || type == FDOUBLE_PTR)
-              ? VPREC_RANGE_BINARY64_DEFAULT
-              : VPREC_RANGE_BINARY32_DEFAULT;
+          (type == FDOUBLE || type == FDOUBLE_PTR) ? function_inst->OpsRange64
+                                                   : function_inst->OpsRange32;
       function_inst->input_args[i].mantissa_length =
-          (type == FDOUBLE || type == FDOUBLE_PTR)
-              ? VPREC_PRECISION_BINARY64_DEFAULT
-              : VPREC_PRECISION_BINARY32_DEFAULT;
+          (type == FDOUBLE || type == FDOUBLE_PTR) ? function_inst->OpsPrec64
+                                                   : function_inst->OpsPrec32;
     }
 
     if (type == FDOUBLE) {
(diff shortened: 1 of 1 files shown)


roblim1:
Confirming that the output remains the same for `test_vprec_instrument` with this fix, and that rounding is applied to the targeted argument/variables and not globally.

pablooliveira:
This looks good to me ! @mistoan since it's in vprec I would appreciate your feedback on this before we merge.

Thanks

src/backends/interflop-vprec/interflop_vprec.c
```diff
@@ -777,10 +777,10 @@ void _interflop_enter_function(interflop_function_stack_t *stack, void *context,
     function_inst->isIntrinsicFunction = function_info->isIntrinsicFunction;
     function_inst->useFloat = function_info->useFloat;
     function_inst->useDouble = function_info->useDouble;
-    function_inst->OpsRange64 = VPREC_RANGE_BINARY64_DEFAULT;
-    function_inst->OpsPrec64 = VPREC_PRECISION_BINARY64_DEFAULT;
-    function_inst->OpsRange32 = VPREC_RANGE_BINARY32_DEFAULT;
-    function_inst->OpsPrec32 = VPREC_PRECISION_BINARY32_DEFAULT;
+    function_inst->OpsRange64 = VPRECLIB_BINARY64_RANGE;
```
mistoan, line 780: I don't think this is right.
We're building a profile of the calls, at this point.

So the default precision to be used in this function is that given by the code, not chosen by the user at a specific execution.

src/backends/interflop-vprec/interflop_vprec.c
```diff
@@ -843,13 +843,11 @@ void _interflop_enter_function(interflop_function_stack_t *stack, void *context,
       function_inst->input_args[i].min_range = INT_MAX;
       function_inst->input_args[i].max_range = INT_MIN;
       function_inst->input_args[i].exponent_length =
-          (type == FDOUBLE || type == FDOUBLE_PTR)
-              ? VPREC_RANGE_BINARY64_DEFAULT
-              : VPREC_RANGE_BINARY32_DEFAULT;
+          (type == FDOUBLE || type == FDOUBLE_PTR) ? function_inst->OpsRange64
```
mistoan, line 846: clang-format?

mistoan (changes requested):
I think there's a misunderstanding in the logic behind the initialization of values used for function-instrument.

src/backends/interflop-vprec/interflop_vprec.c
pablooliveira: It seems this has passed the clang-format test in the builder bot though...

src/backends/interflop-vprec/interflop_vprec.c
roblim1: Fair enough.

roblim1: I think there needs to be a more abstract way of handling vprec runs.  If I, as a user, pass in --range-binary64=11 for the virtual precision backend, I would expect that all floating points would be intercepted with this, regardless of whether an inclusive/exclusive file was passed in.

roblim1: In addition, interflop_init already sets the floating point precision at default modes:
  
   https://github.com/verificarlo/verificarlo/blob/1c9f85669a0011a32bca520c2b9972a7479a9bf1/src/backends/interflop-vprec/interflop_vprec.c#L1506

I stand correct in my commit, as it works in my case and what I expect from Verificarlo's vprec backend.

roblim1:
In addition, interflop_init already sets the floating point precision at default modes:

https://github.com/verificarlo/verificarlo/blob/1c9f85669a0011a32bca520c2b9972a7479a9bf1/src/backends/interflop-vprec/interflop_vprec.c#L1506

I stand correct in my commit, as it works in my case and what I expect from Verificarlo's vprec backend.

roblim1:
The default values for floating point values are being set here:

https://github.com/verificarlo/verificarlo/blob/1c9f85669a0011a32bca520c2b9972a7479a9bf1/src/backends/interflop-vprec/interflop_vprec.c#L1506

As a user to the vprec backend, I would expect that by passing  --precision-binary64=11 that I will be getting results in that precision.  If I am misinterpreting the functionality of the virtual precision backend, please inform.  Thanks.

roblim1:
The default values for floating point values are being set here:

https://github.com/verificarlo/verificarlo/blob/0a57a69601073d14b50af4675a78a50af214e698/src/backends/interflop-vprec/interflop_vprec.c#L1510

As a user to the vprec backend, I would expect that by passing --precision-binary64=11 that I will be getting results in that precision. If I am misinterpreting the functionality of the virtual precision backend, please inform.

roblim1:
Otherwise, I stand correct in my commit.

mistoan:
> The default values for floating point values are being set here:
> 
> https://github.com/verificarlo/verificarlo/blob/0a57a69601073d14b50af4675a78a50af214e698/src/backends/interflop-vprec/interflop_vprec.c#L1510
> 
> As a user to the vprec backend, I would expect that by passing --precision-binary64=11 that I will be getting results in that precision. If I am misinterpreting the functionality of the virtual precision backend, please inform.

You are right, the default values are set at the line that you pointed to.
And then they are set to user-set values, when available, here:
https://github.com/verificarlo/verificarlo/blob/0a57a69601073d14b50af4675a78a50af214e698/src/backends/interflop-vprec/interflop_vprec.c#L1281

The code which you modified relates to the function-instrument compilation pass.
This is extra functionality, allowing us to instrument at the level of function calls, not just at the operation level, as we were able to do before.

The lines which you modified set the default values for the precision of function calls, on the profiling run of function-instrument.
I think it's natural to assume that, without any prior information o a given call, the precision at which it should be executed is the one given by the format it uses.
@Avaquet and @pablooliveira  added this functionality, they can correct me if I'm wrong with my assumptions here.

roblim1:
Is parse_opt ever invoked?  In my current run with test_vprec_instrument
  https://github.com/verificarlo/verificarlo/blob/master/tests/test_vprec_instrument/test.sh

I do not see parse_opt ever being invoked.

roblim1:
I see what is happening.  parse_opt gets invoked when --instrument=all gets passed in. 

It is odd that parse_opt does not get invoked in [Eigen FFT](https://github.com/libigl/eigen/blob/master/bench/benchFFT.cpp), but does show up in the other libraries I am evaluating.  The only way I am able to run Verificarlo with the vprec backend on Eigen FFT is with the current PR.  I am able to run other Eigen benchmarks with the default values.

Either way, thanks for the insight!

PR #232, opened 2021-04-25
Updated TargetLibraryInfo call.  Otherwise, library calls aren't inte…

roblim1:
…rcepted.

### DIFF
--- a/src/libvfcfuncinstrument/libVFCFuncInstrument.cpp
+++ b/src/libvfcfuncinstrument/libVFCFuncInstrument.cpp
@@ -408,11 +408,6 @@ struct VfclibFunc : public ModulePass {
       if ((F.getName().str() != "main") && F.size() != 0) {
         OriginalFunctions.push_back(&F);
       }
-    }
-    /*************************************************************************
-     *                  Gather all library calls                             *
-     *************************************************************************/
-    for (auto &F : M) {
 #if LLVM_VERSION_MAJOR >= 10
       TLI = &(getAnalysis<TargetLibraryInfoWrapperPass>()).getTLI(F);
 #else
--- a/src/libvfcfuncinstrument/libVFCFuncInstrument.cpp
+++ b/src/libvfcfuncinstrument/libVFCFuncInstrument.cpp
@@ -384,8 +384,6 @@ struct VfclibFunc : public ModulePass {
   VfclibFunc() : ModulePass(ID) { inst_cpt = 1; }
 
   virtual bool runOnModule(Module &M) {
-    TargetLibraryInfoWrapperPass TLIWP;
-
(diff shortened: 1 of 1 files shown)


mistoan:
> @mistoan this could be related to the missing `sqrt` call you observed on PATMOS analysis with FuncInstrument ?

I'm not sure.
Does this actually provide more up-to-date information?
The pass isn't redone, but I might be missing something.

> 
> @roblim1 I think this looks nice and can be merged. Could you provide a bit of context to document before we merge ? Which particular library call were you missing before this fix ?
> 
> Thanks

roblim1:
Actually, I have another pass that does this detection and will update the PR shortly.  Thanks.

roblim1:
Branch updated.  Also,I think there needs to be logic here to make this work:

https://github.com/verificarlo/verificarlo/blob/1f9d7f22fbb5258aa9fc043da919c90f8df43e79/src/libvfcfuncinstrument/libVFCFuncInstrument.cpp#L539

roblim1:
Branch updated.  Also, I think there needs to be logic here to make this work:

https://github.com/verificarlo/verificarlo/blob/1f9d7f22fbb5258aa9fc043da919c90f8df43e79/src/libvfcfuncinstrument/libVFCFuncInstrument.cpp#L539

roblim1:
Branch updated.  Also,I think there needs to be logic here to make this work:

https://github.com/verificarlo/verificarlo/blob/1f9d7f22fbb5258aa9fc043da919c90f8df43e79/src/libvfcfuncinstrument/libVFCFuncInstrument.cpp#L539

roblim1:
Branch updated.  Also, I think there needs a check here to make this work:

https://github.com/verificarlo/verificarlo/blob/1f9d7f22fbb5258aa9fc043da919c90f8df43e79/src/libvfcfuncinstrument/libVFCFuncInstrument.cpp#L539

roblim1:
Branch updated.  Also, I think there needs to be a check here to make this work:

  https://github.com/verificarlo/verificarlo/blob/1f9d7f22fbb5258aa9fc043da919c90f8df43e79/src/libvfcfuncinstrument/libVFCFuncInstrument.cpp#L539

roblim1:
Library pass done.  I could produce a case where Verificarlo fails with the current implementation with respect to library calls.

pablooliveira:
Thanks @roblim1, could you please provide a short test so that I can test this before merging it ?

roblim1:
Ok, never mind.  The current implementation detects library calls fine.  In my case, I needed to update [haveFloatingPointArithmetic()](https://github.com/verificarlo/verificarlo/blob/master/src/libvfcfuncinstrument/libVFCFuncInstrument.cpp#L70) to detect floats and doubles from derived types. Thanks.

PR #244, opened 2021-06-02
Fix vector instrumentation + minor fixes:

yohanchatelain:
- Vector instrumentation was invalid for unsupported vector:
      * New implementation used directly the signatures of vfcwrapper
      	in the instrumentation pass. To do so, we merge source and
	vfcwrapper IR files befor applying the instrumentation pass.
      * Add test for checking vectorization instrumentation in
      	tests/test_vectorization
    - Fix compilation of libVFCFuncInstrument.cpp for llvm-4.0
    - Update tests:
      * Remove test.sh.in generations for test_bitmask_backend,
      	test_fortran and test_fortran_NAS} and used paths.sh instead
      * Update clean.sh for some tests
      * Update test_branch_instrumentation test to check if scalar and
      	vector instrumentation give the same result
      * Fix test_cpp when using llvm-4


Sholde:
Hi,

I test your code but it seems not working when we don't specify the flags `-march=native` or `-mavx`. Then I think it works only with **AVX** set. But I think we need that woks even if we have only **SSE** set.

I test your implementation on the code below:
```
#include <stdio.h>

typedef float float16 __attribute__((ext_vector_type(16)));

int main()
{
  float16 a;
  float16 b;

  for (int i = 0; i < 16; i++)
    {
      a[i] = 8.7565;
      b[i] = 1.1343;
    }
  
  float16 c = a + b;

  for (int i = 0; i < 16; i++)
    printf("c[i] = %f\n", c[i]);
  
  return 0;
}
```

I have this output **without AVX**:
```
$ verificarlo-c main.c -o main
$ ./main 
Info [verificarlo]: loaded backend libinterflop_ieee.so
c[i] = 9.890800
c[i] = 9.890800
c[i] = 9.890800
c[i] = 9.890800
c[i] = 9.890800
c[i] = 9.890800
c[i] = 9.890800
c[i] = 9.890800
c[i] = 9.890800
c[i] = 9.890800
c[i] = 9.890800
c[i] = 9.890800
c[i] = 8.756500
c[i] = 8.756500
c[i] = 8.756500
c[i] = 8.756500
```

And I have this output **with**:
```
$ verificarlo-c main.c -o main -mavx
$ ./main 
Info [verificarlo]: loaded backend libinterflop_ieee.so
c[i] = 9.890800
c[i] = 9.890800
c[i] = 9.890800
c[i] = 9.890800
c[i] = 9.890800
c[i] = 9.890800
c[i] = 9.890800
c[i] = 9.890800
c[i] = 9.890800
c[i] = 9.890800
c[i] = 9.890800
c[i] = 9.890800
c[i] = 9.890800
c[i] = 9.890800
c[i] = 9.890800
c[i] = 9.890800
```

**cpuinfo**:
```
processor    : 0
vendor_id    : GenuineIntel
cpu family    : 6
model        : 158
model name    : Intel(R) Core(TM) i5-8400 CPU @ 2.80GHz
stepping    : 10
cpu MHz        : 2807.996
cache size    : 9216 KB
physical id    : 0
siblings    : 4
core id        : 0
cpu cores    : 4
apicid        : 0
initial apicid    : 0
fpu        : yes
fpu_exception    : yes
cpuid level    : 22
wp        : yes
flags        : fpu vme de pse tsc msr pae mce cx8 apic sep mtrr pge mca cmov pat pse36 clflush mmx fxsr sse sse2 ht syscall nx rdtscp lm constant_tsc rep_good nopl xtopology nonstop_tsc cpuid tsc_known_freq pni pclmulqdq ssse3 cx16 pcid sse4_1 sse4_2 x2apic movbe popcnt xsave avx rdrand hypervisor lahf_lm abm 3dnowprefetch invpcid_single pti fsgsbase avx2 invpcid rdseed clflushopt md_clear flush_l1d
bugs        : cpu_meltdown spectre_v1 spectre_v2 spec_store_bypass l1tf mds swapgs itlb_multihit srbds
bogomips    : 5615.99
clflush size    : 64
cache_alignment    : 64
address sizes    : 39 bits physical, 48 bits virtual
power management:

```

yohanchatelain:
Thanks @Sholde for pointing out. Actually the issue is that I passed `-march=native` to compile `vfcwrapper.c` which always use the more recent ISA. I should check if `-march` or `-m<isa>` is passed in options to have the same for compiling `vfcwrapper.c`

Sholde:
It works for me.

src/vfcwrapper/main.c
```diff
@@ -51,25 +51,45 @@
...
+// typedef float float2 __attribute__((ext_vector_type(2)));
+// typedef float float4 __attribute__((ext_vector_type(4)));
+// typedef float float8 __attribute__((ext_vector_type(8)));
+// typedef float float16 __attribute__((ext_vector_type(16)));
+
+// typedef double double2 __attribute__((ext_vector_type(2)));
+// typedef double double4 __attribute__((ext_vector_type(4)));
+// typedef double double8 __attribute__((ext_vector_type(8)));
+// typedef double double16 __attribute__((ext_vector_type(16)));
+
```
Sholde, line 68: Can we remove these comment lines ?
As we demand to user to use only `clang` or `gcc` and `clang` using `OpenCL` or `gcc` vectors types.
We might as well use `gcc` vectors.
Because we don't use `OpenCL` tricks like `v.xyz`.

yohanchatelain: Right, I forgot to remove these lines. Actually I'll change for `ext_vector_type` which is the OpenCL extension.

src/clang_options_parser/clang_options_parser.py
```diff
@@ -0,0 +1,228 @@
...
+#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the               *
+#  GNU General Public License for more details.                                *
+#                                                                              *
+#  You should have received a copy of the GNU General Public License           *
+#  along with Verificarlo.  If not, see <http://www.gnu.org/licenses/>.        *
+#                                                                              *
+# *******************************************************************************
+
+
+target_specific_options_x86 = set([
```
titeup, line 26: is this list taken from somewhere? if yes that would be nice at least in comment to cite the source. And check the license if relevant.

titeup (commented):
not much to say.

src/clang_options_parser/clang_options_parser.py
yohanchatelain: It was taken from `gcc --help-target`. Normally, gcc flags are compatible with clang but we could miss some. I would like to have an automate process for handling those options.

yohanchatelain:
> Hi,
> 
> Sorry but I retested your code with `-march=native` but it doesn't work. (I tested before only with `-mavx`).
> I think is due to your list which include only `-march=`, doesn't include all values of this options.
> And it is not the only.
> 
> Maybe you will need to add all values in the list if you don't find another options to resolve that.
> 
> **UPDATE:**
> Furthermore if we put an invalid option (like `-march=native` for the moment) we have a python traceback.
> Maybe we can add a better error message ?

It's surprising because the option parser does not look at which value is taken but only if the option is present. So it should accept any values. Could you show the traceback?

Sholde:
Sorry forget what I said. I didn't update my python environment variable with my new version of python xD
You are right. It works.

titeup (commented):
Any news outside github from @pablooliveira about this list vs. license?
I should meet him for other matters tomorrow morning, if it is not solved already we can try to take a few minutes to discuss that comment.

src/clang_options_parser/clang_options_parser.py
pablooliveira: Can we get the same list from the LLVM project, instead of taking a GPL header file ? I see no other way around.

pablooliveira (changes requested):

pablooliveira:
By the way this is really nice work @yohanchatelain. Thanks for tackling this tricky issue !

src/clang_options_parser/clang_options_parser.py
yohanchatelain: Ok I found an other way to solve the issue without parsing the inputs flags so we can remove this file. I'm just not sure if it preserves the optimization flags so we should discuss about it. The idea is to check if caller (source code) and callee (vfcwrapper) functions have the same ABI, meaning if they have the same `target-cpu` and `target-features` attributes. If not, I update the caller with the atttributes' callee. In that case, I always compile vfcwrapper with `-03 -march=native` to trigger vectorization. My main concern is if promoting the caller function may trigger optimization that are not specified by the user. Let's say it passes `-O0` but we optimize the function anyway. However, this case should not met in practice because passing -O0 as input flags should not trigger vectorization. I'll push my update soon.

yohanchatelain: To precise why the previous solution was not working, within fuzzy the optimization flags used for compiling the Python Interpreter and those for the numpy packages are different. So the definition of the vectorized functions (`.vfcwrapper.o`) are different and the definition of the interpreter are taken instead of those of numpy. Causing to pass values in the wrong registers (xmm instead of zmm for avx512f). By always using the same optimization flags to compile the `vfcwrapper.c` we ensure that definitions are consistent across different library compilation.

pablooliveira:
Thanks Yohan, the new design to detect the ABI difference between backend and IR looks good. Since this is a big PR I did not have time to review it fully, can you wait a couple of days for it to be reviewed or is the merge needed quick on your side ?

yohanchatelain:
> Thanks Yohan, the new design to detect the ABI difference between backend and IR looks good. Since this is a big PR I did not have time to review it fully, can you wait a couple of days for it to be reviewed or is the merge needed quick on your side ?

Thanks @pablooliveira , sure we can for it. It also would be nice to make a new version after merging this PR because it fixes an important bug, specially for fuzzy.

pablooliveira (approved):
This looks very nice ! Thanks :tada:

titeup (approved):

titeup:
ok to merge. Congrats and thanks @yohanchatelain, that was not an easy issue to sort out!

Issue #245, opened 2021-06-04
Warning when using math library


Sholde:
Hi,

Last month I wrote a code which using **C math library** and I found a little bug. 

When I didn't link the library I have the normal error due to of undefined reference. But when I linked it on verificarlo command line, `clang` tall me that is `unused`. But it is necessary due to the call of `sqrt` functions.

The code:
```c
#include <math.h>

int main()
{
  sqrt(4);
}
```

Error **without** link:
```shell
$ verificarlo-c sqrt.c
/usr/bin/ld: sqrt.o: in function `main':
sqrt.c:(.text+0x16): undefined reference to `sqrt'
clang-11: error: linker command failed with exit code 1 (use -v to see invocation)
/usr/local/bin/verificarlo: command failed:
/usr/bin/clang @/tmp/tmpw83jhidz
```

Warning **with** link:
```shell
$ verificarlo-c sqrt.c -lm
clang-11: warning: -lm: 'linker' input unused [-Wunused-command-line-argument]
clang-11: warning: -lm: 'linker' input unused [-Wunused-command-line-argument]
```

Using `llvm 10` and `clang 11`.

yohanchatelain:
This is a normal behavior, all flags are passed to the compiler and the liker but at compile stage we don't need libraries flags. We could parse them and dispatch them correctly.

Sholde:
Yes, when I said it is a little bug I mean that is not the same behaviour that clang because we have this warning.

PR #248, opened 2021-06-09
separate the library flags to activate them only in the linker stage

Sholde:
Hi,

I make these changes to solve the issue #245.

### DIFF
--- a/verificarlo.in.in
+++ b/verificarlo.in.in
@@ -97,6 +97,7 @@ def shell_escape(argument):
 def parse_extra_args(args):
     sources = []
     options = []
+    libraries = []
 
     for a in args:
         if is_fortran(a):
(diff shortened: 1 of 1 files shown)


titeup (commented):
Looks good to me, but I would appreciate a double check by @yohanchatelain or @pablooliveira  , to make sure we are not braking complex building that has been done, in particular in the fuzzy docker image of @gkiar.
Especially since it is intended to remove a potentially misleading but armless warning not fix a critical bug.

pablooliveira:
This looks nice and should work well. I'm only concerned about possible bugs when linking a mix of dynamic "-l" and static ".a" libraries; the order could be changed in your commit. But I think this should still be fine to merge, if the bugs triggers we can always fix this at that point.

pablooliveira (approved):

PR #252, opened 2021-06-14
Sparse perturbations

gkiar:
Hi!

I am in the process of adding a `--sparsity` flag to the verificarlo MCA backend, and thought it was time to get feedback :)

The purpose of this modification is to enable the study of perturbation density without the additional confounding variable of _which_ operations are actually being perturbed. This could potentially have implications for adopting MCA on large workflows. For example, if perturbing even 1 in 2 operations allows us to capture similar variability to fully dense MCA, it would save resource consumption by half.

To introduce this, I have modified the `interflop_mca.c` backend file as follows:
- added a variable which can be parsed by the backend to store the new hyperparameter (`--sparsity`, `-n`)
- a function, `_mca_skip_eval`, which checks whether a perturbation should be skipped. This is determined randomly rather than ever N operations so that there is no bias based on the sequence of operations in a given program.
- a branch condition in the `_INEXACT` macro function which uses the above to potentially skip perturbations. If the sparsity is 1 (no intended skipping) then this branch is just an equality test and doesn't bother generating a random number to reduce overhead.

The tests pass on my local build, but I have yet to add a new set of tests for this functionality, and would appreciate input on what you'd like to see as an effective test for this. Thanks!


titeup:
Hi @gkiar , funny idea. However quite different from pure MCA even if it uses it. I wonder if it cannot be a backend on its own, eventually calling MCA backend. I am thinking of it because of the additional overhead of a call (inlined?) and test before each operation even in the 'dense' case.
Did you measure it?

gkiar:
Hi @titeup, I just pushed a first "test" that measures timing performance across a series of runs using increasing sparsity. The test ran the script of 1000 additions 100x for each sparsity. I imagine caching is skewing the timing a bit, but here are a set of results when run on my system (times in nano seconds):

```
Checking backend MCA at SPARSITY 1
85506144

Checking backend MCA at SPARSITY 2
73715243

Checking backend MCA at SPARSITY 5
66718159

Checking backend MCA at SPARSITY 10
61080513

Checking backend MCA at SPARSITY 20
59402456

Checking backend MCA at SPARSITY 100
64390430

Checking backend IEEE backend
55765014
```

gkiar:
With help from @yohanchatelain I've updated the test, and now run it against both the modified MCA backend and the original (in the current `master` branch).

The modified MCA backend:
```
# ./test.sh 
Sparsity: 1 | 0.0% IEEE (Time Elapsed: 525 ms)
Sparsity: 2 | 50.031% IEEE (Time Elapsed: 527 ms)
Sparsity: 4 | 75.151% IEEE (Time Elapsed: 433 ms)
Sparsity: 10 | 89.92% IEEE (Time Elapsed: 420 ms)
Sparsity: 100 | 99.01% IEEE (Time Elapsed: 418 ms)
```

The previous MCA backend (there is no difference across rows, in this setting):
```
# ./test.sh 
Sparsity: 1 | 0.0% IEEE (Time Elapsed: 520 ms)
Sparsity: 2 | 0.0% IEEE (Time Elapsed: 518 ms)
Sparsity: 4 | 0.0% IEEE (Time Elapsed: 550 ms)
Sparsity: 10 | 0.0% IEEE (Time Elapsed: 476 ms)
Sparsity: 100 | 0.0% IEEE (Time Elapsed: 496 ms)
```

@titeup, what do you think? From this small example, it looks like the overhead is smaller than the between-run variability to me.

titeup:
Looks good that it decrease significantly with sparsity. However it is missing one point in the experiment with is the current performance of the MCA backend on master (compare to sparsity ==1) to see if the additional test is impacting perf.

gkiar:
Hey @titeup — that performance is shown in this comment: https://github.com/verificarlo/verificarlo/pull/252#issuecomment-860856989

titeup:
> Hey @titeup — that performance is shown in this comment: [#252 (comment)](https://github.com/verificarlo/verificarlo/pull/252#issuecomment-860856989)

right i missread it. Can you do 5 execution of sparsity=1 with the new code to check we are in the same interval?

I suppose you are using the sparsitytest from you PR for the measuement. Can you also measure the difference with fp32 ?
With double, the quad computation are quite heavy and can easily mask the additional overhead, not sure how it is compare to float input that we used native double for MCA.

If everything right I will start reviewing the code. But at first sight it looks good and should not cause any trouble.

gkiar:
Hi @titeup, here are the results from the 5 runs with `--sparsity=1`:
```
DOUBLE
  Sparsity: 1 | 0.0% IEEE (Time Elapsed: 533 ms)
  Sparsity: 1 | 0.0% IEEE (Time Elapsed: 551 ms)
  Sparsity: 1 | 0.0% IEEE (Time Elapsed: 548 ms)
  Sparsity: 1 | 0.0% IEEE (Time Elapsed: 552 ms)
  Sparsity: 1 | 0.0% IEEE (Time Elapsed: 537 ms)
```

So, the `master` backend ranged from `476–550` ms vs `533–552` ms with sparsity added for this test, with N=5 each.

When using fp32, the profiling was as follows:

***Sparse backend, various sparsities***:
```
FLOAT
  Sparsity: 1 | 0.0% IEEE (Time Elapsed: 511 ms)
  Sparsity: 2 | 50.063% IEEE (Time Elapsed: 520 ms)
  Sparsity: 4 | 75.139% IEEE (Time Elapsed: 529 ms)
  Sparsity: 10 | 89.955% IEEE (Time Elapsed: 530 ms)
  Sparsity: 100 | 99.01% IEEE (Time Elapsed: 413 ms)
```

***Sparse backend, various runs, sparsity=1***:
```
FLOAT
  Sparsity: 1 | 0.0% IEEE (Time Elapsed: 511 ms)
  Sparsity: 1 | 0.0% IEEE (Time Elapsed: 523 ms)
  Sparsity: 1 | 0.0% IEEE (Time Elapsed: 513 ms)
  Sparsity: 1 | 0.0% IEEE (Time Elapsed: 514 ms)
  Sparsity: 1 | 0.0% IEEE (Time Elapsed: 514 ms)
```

***`master` backend, various runs***:
```
FLOAT
  Sparsity: 1 | 0.0% IEEE (Time Elapsed: 511 ms)
  Sparsity: 2 | 0.0% IEEE (Time Elapsed: 511 ms)
  Sparsity: 4 | 0.0% IEEE (Time Elapsed: 511 ms)
  Sparsity: 10 | 0.0% IEEE (Time Elapsed: 501 ms)
  Sparsity: 100 | 0.0% IEEE (Time Elapsed: 489 ms)
```

From these results, the range also looks similar across the floats: `master` ranged from `489–511` and sparse=1 ranged from `511–523`, for N=5 each.

pablooliveira:
That is a very nice idea. I think sampling randomly which operations to perturb is a great addition.

src/backends/interflop-mca/interflop_mca.c
```diff
@@ -186,6 +189,16 @@ static double _mca_rand(void) {
...
 }
 
+static bool _mca_skip_eval(const int sparsity) {
+  /* Returns a bool for determining whether an operation should skip */
+  /* perturbation. false -> perturb; true -> skip. */
+  if (sparsity == 1) {
+    return false;
+  }
+  /* e.g. for sparsity=10, all random values > 0.1 = true -> no MCA*/
+  return (_mca_rand() > (1.0 / sparsity));
```
titeup, line 199: why not using an integer rand, avoid the div and cast(s) and compute the sparsity threshold onece for all as a fraction of the unit64 (or 32 64 might be overkill) and compare to this threshold.

src/backends/interflop-mca/interflop_mca.c
```diff
@@ -186,6 +189,16 @@ static double _mca_rand(void) {
   return tinymt64_generate_doubleOO(&random_state);
 }
 
+static bool _mca_skip_eval(const int sparsity) {
```
titeup, line 192: inline ?

titeup (changes requested):
Everything looks fine. Only one small suggestion to reduce the overhead of the call and associated test for sparsity. Open to discussion, it can also be ok as it is since it does not influence the path without sparsity.

src/backends/interflop-mca/interflop_mca.c
gkiar:
```suggestion
  return (rand() % sparsity);
```

Hi @titeup, would the following work? In this case, only when the mod is zero would a perturbation take place. If we do this, should I initialize the random state with `srand` next to when we initialize the other on line 590?

src/backends/interflop-mca/interflop_mca.c
gkiar: Hi @titeup , I'm not sure I understand what you mean here. Sorry about that.

src/backends/interflop-mca/interflop_mca.c
titeup: the 2 random generator (mac_rand and rand()) are independent and therefore the rand seeding has no relationship to mca_rand. However we implemented an independent mca_rand as Mersen Twister in vericarlo to make sure that we are not interfering with application that might use rand(). I believe we are in the same situation.
What could be done is use our MersenTwister output as int64 and do the modulo. There are This would work.  Look in common/tinymt64.h.  I believe  tinymt64_generate_uint64 is what you are looking for.

src/backends/interflop-mca/interflop_mca.c
titeup: static inline bool ...
To instruct the compiler to inline your call and avoid call overhead. This is useful since backends are not using aggressive compiler options to not mess the very sensitive (and difficult to debug) arithmetic we are using.

src/backends/interflop-mca/interflop_mca.c
gkiar: I'm realizing that `rand` is biased, but I think this should work with a uniform random number generator. Do you have a recommendation?

src/backends/interflop-mca/interflop_mca.c
gkiar:
```suggestion
static inline bool _mca_skip_eval(const int sparsity) {
```

gkiar: Thanks!

src/backends/interflop-mca/interflop_mca.c
gkiar: @titeup perfect, thanks. When evaluating with the modulo, I notice that there is some bias and across sparsity values we are not consistently within a percent or two of what we expect. I tried implementing your suggestion: pre-computing the threshold as (`UINT64_MAX/sparsity`)  and continue with a `>` comparison, but also noticed more bias in this, as far as when sparsity was 4 we still perturbed 29% of operations instead of 25% over a few hundred thousand runs. Should we try harder to make this uniform, or should I just precompute the threshold to save the division each iteration and keep the logic otherwise the same?

yohanchatelain: One solution to avoid the division is to pass the sparsity as a percentage ratio instead of an integer, like `10 -> 0.1`.

gkiar: I ended up going with @yohanchatelain 's recommendation, which also gives us a lot of flexibility in controlling the sparsity at a finer resolution.

src/backends/interflop-mca/interflop_mca.c
```diff
@@ -479,6 +496,18 @@ error_t parse_opt(int key, char *arg, struct argp_state *state) {
     /* flush-to-zero */
     ctx->ftz = true;
     break;
+  case KEY_SPARSITY:
+    /* sparse perturbations */
+    errno = 0;
+    ctx->sparsity = strtod(arg, &endptr);
```
yohanchatelain, line 502: Use `strtof` instead of `strtod` since `sparsity` is a `float`.

src/backends/interflop-mca/interflop_mca.c
```diff
@@ -495,32 +524,35 @@ void init_context(t_context *ctx) {
   ctx->daz = false;
   ctx->ftz = false;
   ctx->seed = 0ULL;
+  ctx->sparsity = 1;
```
yohanchatelain, line 527: `1` -> `1.0f`

src/backends/interflop-mca/interflop_mca.c
```diff
@@ -495,32 +524,35 @@ void init_context(t_context *ctx) {
...
-      ctx->daz ? "true" : "false", key_ftz_str, ctx->ftz ? "true" : "false");
+  logger_info("load backend with "
+              "%s = %d, "
+              "%s = %d, "
+              "%s = %s, "
+              "%s = %s, "
+              "%s = %d, "
+              "%s = %s, "
+              "%s = %s and "
+              "%s = %d"
```
yohanchatelain, line 541: `%s = %d` -> `%s = %f`

src/backends/interflop-mca/interflop_mca.c
```diff
@@ -186,6 +189,16 @@ static double _mca_rand(void) {
   return tinymt64_generate_doubleOO(&random_state);
 }
 
+static inline bool _mca_skip_eval(const float sparsity) {
+  /* Returns a bool for determining whether an operation should skip */
+  /* perturbation. false -> perturb; true -> skip. */
+  if (sparsity >= 1) {
```
yohanchatelain, line 195: `1` -> `1.0f`

yohanchatelain (approved):
LGTM! Good contribution :slightly_smiling_face:

titeup (approved):
looks ok, approved!

gkiar:
Great! Thanks for the help, @titeup and @yohanchatelain ! @pablooliveira, do you have any comments before we merge?

pablooliveira:
No this looks great ! Congrats.

Issue #261, opened 2021-06-16
In trouble with python-git


Sholde:
Hello,

Recently, `verificarlo` added the `python-git` dependency. 
So I installed it with `pip`. 
But the configuration file failed to reach it. 
So I consulted the online documentation https://pypi.org/project/python-git/.
But they say that we have to run a few lines to set up an initial configuration.
I don't really understand where we need to configure it, in the directory where we work on `verificarlo` or where we build it?

Another question that's been on my mind, if we have to configure it manually on `verificarlo` then isn't it automatically?

Thank you,

PurplePachyderm:
Hi,

There are actually a few Python modules available to interact with Git, including `python-git` (https://pypi.org/project/python-git/) as you pointed out, and `GitPython` (https://gitpython.readthedocs.io/en/stable/tutorial.html). This can be quite confusing, but `GitPython` is actually the one used in Verificarlo.

I just noticed that the documentation has not been updated with the new Python dependencies so I'll probably fix this in a PR to avoid any confusions in the future.

Sholde:
Thank you.

PR #262, opened 2021-06-17
Update documentation with new Pyton dependencies

PurplePachyderm:
As pointed out in #261, the documentation was not up-to-date with the
newest Python dependencies that have been added. This new version lists all
the required Python packages and updates the command in the installation
example.


pablooliveira (approved):
Thanks ! Our dependencies are getting out of hand :-) We will have to think about how to manage them in the future.

PurplePachyderm:
Maybe a good start would be to add a `requirements.txt` at some point to at least manage the Python dependencies.

PR #263, opened 2021-06-17
removing the clean of tests/paths.sh file

Sholde:
This file was generated by the configuration script before. But now it
have a default setting.

### DIFF
--- a/tests/clean.sh
+++ b/tests/clean.sh
@@ -16,4 +16,4 @@ for t in $tests ; do
     fi
 done
 
-rm -Rf *~ testplan.log testplan.trs test-suite.log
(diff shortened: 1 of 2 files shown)


Sholde:
But this default setting depends on the architecture?
For example me, I don't have `llvm-7`, so this setting is obsolete for me?

yohanchatelain:
I think we should remove `tests/path.sh` file from the tracked file because we don't have default setting. Maybe it has been tracked by mistake. This file must be generated by the configure.

Sholde:
Agreed. I think I will make another PR.

yohanchatelain:
You can just untracked the file and push a new commit.

yohanchatelain (approved):

Issue #264, opened 2021-06-17
Release v0.4.3?


yohanchatelain:
Can we release a new version? since major contributions have been added since the last release

Sholde:
I think we should merge #263 (or fix the issue) first in order to have a clean behavior for the `make clean` command.

yohanchatelain:
I think that it does not impact the behaviour of the backends neither the general mechanic. Moreover, it is just an issue with a tracked file that should not be.

PR #265, opened 2021-06-17
Pre-release v0.4.3 (#264)

yohanchatelain:


pablooliveira (approved):
This look complete. Thanks for updating the CHANGELOG !

I think it would be good practice to update the CHANGELOG, step by step for each PR merged. I know that is what @yohanchatelain has already being doing; but I think it would be nice if we all try to do it in the future.

pablooliveira (changes requested):
Actually before you merge, I think there are a few big changes and improvements in this release.

Maybe we should increase the major version number since this is clearly more than an incremental release. I think we could have v0.5.0 instead.

Moreover the CI workflow tool from @PurplePachyderm is quite a big project; maybe we could have a slightly longer description to help the user discover it. I would suggest:

* Add vfc_ci, a new tool which integrates verificarlo in a CI workflow and generates interactive reports that track numerical errors across the project lifetime.

pablooliveira:
Sorry @yohanchatelain, I added my request for changes a bit late.

yohanchatelain:
Sorry I already merged the PR but I changed the release state to pre-release so we can still modify it.

yohanchatelain:
Btw, you could enable the discussion sections on verificarlo like I did in [sigdigits](https://github.com/verificarlo/significantdigits/discussions), to have a dedicated space to discuss without opening issues.

titeup:
I agree with Pablo comment, 0.5 is preferable.
I even think that the project start to be mature enough to reach 1.0 ;) (with relicensing)
Also there have been some addition in the documentation (ci in particular) so I would add this bullet in the addition (and maybe fused the changed in the doc in the same bullet on top)
I cannot see Matei's fix/changes in func instr and vprec-libm exploration in your list but maybe he did not push it on master yet. I will check with him. It would be nice to have it release so he can claim it in his funded project review.

PR #274, opened 2021-07-30
Fix bug that would cause vfc_ci test to fail on warning

PurplePachyderm:
When a warning was thrown while reading the probes' CSV file, the `read_probes_csv` function would return incorrect values, which would cause the entire command to fail. This commit fixes this issue so that warnings can be correctly handled.


pablooliveira (approved):
LGTM !

PR #277, opened 2021-08-31
Parallel noise generation for the MCA back-end -- OpenMP specific version

mistoan:
This pull-request adds an option for the MCA back-end to use a RNG that can be used for programs that use OpenMP parallelism.
For now, the options for the RNGs are standard C re-entrant generators, but more to follow.


yohanchatelain:
Hi @mistoan, thanks for this helpful contribution. I have few questions:
- Why do you use stdlibc RNG and not the tinymt library? Because the purpose of using a separate RNG was to not interfere with the standard ones
- Would it be simpler to make mca-backend and tinymt thread-safe by modifying their global variables to make them thread-safe? By using the `thread_local` attribute for example

pablooliveira:
Hi @yohanchatelain, thanks for the feedback. We were indeed discussing this work with @mistoan in our weekly meeting.

- Regarding the second point, indeed, we want to do a runtime agnostic version. The idea is to use TLS through the `__thread` annotation and avoid any OpenMP construct. We keep this PR open for now just so we have a working prototype.

- Regarding the use of standard RNG. We wanted a quick reentrant RNG to do a first proof of concept. In the final version we were planning to use something such as https://www.pcg-random.org/index.html

mistoan:
> Hi @mistoan, thanks for this helpful contribution. I have few questions:

Hi @yohanchatelain ,

> 
>     * Why do you use stdlibc RNG and not the tinymt library? Because the purpose of using a separate RNG was to not interfere with the standard ones

The problem with the tinymt implementation that is currently used in the MCA backend is that it is not re-entrant.
I chose some standard C implementation for the RNG to get a rapid prototype going.
If you have a look at the code, there's provisions to use different RNGs, which we can test out later.

> 
>     * Would it be simpler to make mca-backend and tinymt thread-safe by modifying their global variables to make them thread-safe? By using the `thread_local` attribute for example

It is, yes.
I think there's parallel implementations for Mersenne Twister, as @pablooliveira has suggested in our meetings.
There's also PCG-random, that @pablooliveira suggested in his reply.

There's another pull-request coming that has a generic implementation, using the `__thread` annotation.

yohanchatelain:
> > Hi @mistoan, thanks for this helpful contribution. I have few questions:
> 
> Hi @yohanchatelain ,
> 
> > ```
> > * Why do you use stdlibc RNG and not the tinymt library? Because the purpose of using a separate RNG was to not interfere with the standard ones
> > ```
> 
> The problem with the tinymt implementation that is currently used in the MCA backend is that it is not re-entrant.
> I chose some standard C implementation for the RNG to get a rapid prototype going.
> If you have a look at the code, there's provisions to use different RNGs, which we can test out later.
> 
> > ```
> > * Would it be simpler to make mca-backend and tinymt thread-safe by modifying their global variables to make them thread-safe? By using the `thread_local` attribute for example
> > ```
> 
> It is, yes.
> I think there's parallel implementations for Mersenne Twister, as @pablooliveira has suggested in our meetings.
> There's also PCG-random, that @pablooliveira suggested in his reply.
> 
> There's another pull-request coming that has a generic implementation, using the `__thread` annotation.

Actually, I'm not sure that tinymt is not re-entrant. Indeed, tinymt does not used internal global states, the only state used is the `tinymt64_t` structure but it is passed by pointer to each functions using it. Moreover, tinymt has no depedencies to external libraries that could make it not thread-safe. The only precaution for me is to have those states thread_local in the mca backend to avoid interference and to carefully select seeds to avoid duplicates and unintended correlation between their outputs. But there is certainly something that I must miss.

Nevertheless, I agree that we should move to PCG RNG.

mistoan:
> > > Hi @mistoan, thanks for this helpful contribution. I have few questions:
> > 
> > 
> > Hi @yohanchatelain ,
> > > ```
> > > * Why do you use stdlibc RNG and not the tinymt library? Because the purpose of using a separate RNG was to not interfere with the standard ones
> > > ```
> > 
> > 
> > The problem with the tinymt implementation that is currently used in the MCA backend is that it is not re-entrant.
> > I chose some standard C implementation for the RNG to get a rapid prototype going.
> > If you have a look at the code, there's provisions to use different RNGs, which we can test out later.
> > > ```
> > > * Would it be simpler to make mca-backend and tinymt thread-safe by modifying their global variables to make them thread-safe? By using the `thread_local` attribute for example
> > > ```
> > 
> > 
> > It is, yes.
> > I think there's parallel implementations for Mersenne Twister, as @pablooliveira has suggested in our meetings.
> > There's also PCG-random, that @pablooliveira suggested in his reply.
> > There's another pull-request coming that has a generic implementation, using the `__thread` annotation.
> 
> Actually, I'm not sure that tinymt is not re-entrant. Indeed, tinymt does not used internal global states, the only state used is the >`tinymt64_t` structure but it is passed by pointer to each functions using it. Moreover, tinymt has no depedencies to external l>ibraries that could make it not thread-safe. The only precaution for me is to have those states thread_local in the mca backend

I don't think this the case, with the current implementation.

> to avoid interference and to carefully select seeds to avoid duplicates and unintended correlation between their outputs.

This isn't done at the moment, either.

> But >there is certainly something that I must miss.

I don't think you're missing anything.
But, in my opinion, these little details that you've pointed to are what this pull request target.

> 
> Nevertheless, I agree that we should move to PCG RNG.

PCG seems like a good solution.
Then again, it comes with the disadvantage of having to rely on an external project as a dependency.

pablooliveira:
Closing this PR which is superseded by #281.

PR #279, opened 2021-09-02
testplan.test should preserve environment when running tests

pablooliveira:
When using custom path --prefix installations, tests broke because the local PATH and PYTHONPATH environment were not preserved. To fix this avoid running tests inside a log shell as this does not preserve the environment.


yohanchatelain (approved):
LGTM, good catch!

PR #280, opened 2021-09-07
Fix name display issue in vfc_ci report

PurplePachyderm:
As explained in 2ef47d2 :

Run names older than one month in the report used to not be rounded as
expected. This caused the name to show the number of elapsed month as a
float. -> Simply fixed by ading the missing call to int.


pablooliveira:
Good catch ! LGTM

pablooliveira (approved):

PR #287, opened 2021-12-09
Mca integer implementation

mistoan:
This pull request brings a version of the MCA back-end where the operations regarding the noise are done using **integer and bitwise** operations.
The only floating-point operations are for the final rounding to the output format.

**Current status:** work-in-progress

**TODO:**
 - [] optimization of the quad part
 - [] after discussions with @titeup, the quad part could be optimized towards the use of a double-precision float and a 64-bit integer, instead of two double-precision floats
 - [] possible optimization:
   - `drand48` only generates 32-bit integers, when generating an integer
   - for now, we therefore need 2 calls for handling floats and 4 calls for handling doubles
   - however, for instrumenting floats we only need the second random number when working on virtual precision < 20; for instrumenting doubles we only need the third and fourth random numbers when working on virtual precision < 48 and the fourth random number when working on virtual precision < 16


yohanchatelain:
HI @mistoan, looks very nice! Is your PR still in the draft? If so, you should convert it to a draft, and if not, I'd like to review it.

titeup:
@mistoan can you share on this PR the perf improvement you are observing so far with this version?
Thanks, 
Eric

pablooliveira:
@mistoan. This PR contains many changes that are unrelated to the MCA computation. Could you please do a clean rebase on master ?

pablooliveira:
Congrats on removing the `if` construct for positive and negative noise. The new version with signed integer noise should be faster I expect. I am also interested in the perf analysis :-)

mistoan:
> > **Current status:** work-in-progress
> > HI @mistoan, looks very nice! Is your PR still in the WIP status? If so, you should convert it to a draft, and if not, I'd like to review it.

Indeed, it's still a work in progress.
I've converted the pull-request to a draft, as you suggested.

mistoan:
> @mistoan can you share on this PR the perf improvement you are observing so far with this version? Thanks, Eric

I've done some quick tests using the methodology discussed [here](https://github.com/verificarlo/verificarlo/discussions/284#discussioncomment-1474997).
I'm testing with 12 threads.
original MCA is at 26.052541s execution time
the integer MCA is at 19.719286s execution time.
Tests is simple precision.

So, for the moment, it seems to be slightly faster.

src/backends/interflop-mca-int/interflop_mca_int.c
```diff
@@ -0,0 +1,681 @@
...
+  // generate a new random 64-bit integer
+  noise = _get_rand_uint64(rng_state, &global_tid_lock, &global_tid);
+
+  // extract the MSB of the noise
+  //  first extract the MSB
+  noise_msb = noise & (((uint64_t)1) << (DOUBLE_EXP_SIZE + DOUBLE_PMAN_SIZE));
+  //  next, shift the MSB all the way to the LSB position
+  noise_msb = noise_msb >> (DOUBLE_EXP_SIZE + DOUBLE_PMAN_SIZE);
+  // noise sign mask is used to create the one's complement of the masked noise
+  noise_sign_mask = (uint64_t)DOUBLE_MASK_ONE + (noise_msb ^ 1);
```
pablooliveira, line 226: I do not think we need to treat separately the sign bit.
Indeed the first bit of `noise` can act as a sign bit naturally.
When shifting the noise with >> (DOUBLE_EXP_SIZE + DOUBLE_PMAN_SIZE) the sign bit will be correctly extended.

It seems a quite simpler implementation is possible that only does:
```c
noise = _get_rand_uint64(rng_state, &global_tid_lock, &global_tid);
noise= noise >> (DOUBLE_EXP_SIZE + DOUBLE_PMAN_SIZE);
x_b64.u64 = x_b64.u64 + noise;
```
I am testing this on a separate test harness to check that it matches SR expected distributions on some simple inexact values and it works well.

pablooliveira (changes requested):

src/backends/interflop-mca-int/interflop_mca_int.c
```diff
@@ -0,0 +1,681 @@
...
+  noise_msb =
+      noise_high & (((uint64_t)1) << (DOUBLE_EXP_SIZE + DOUBLE_PMAN_SIZE));
+  //  next, shift the MSB all the way to the LSB position
+  noise_msb = noise_msb >> (DOUBLE_EXP_SIZE + DOUBLE_PMAN_SIZE);
+  // noise sign mask is used to create the one's complement of the masked noise
+  noise_sign_mask = (uint64_t)DOUBLE_MASK_ONE + (noise_msb ^ 1);
+
+  // apply the mask to the noise, to only keep the noise at the correct
+  // magnitude
+  noise_low &= noise_mask_low;
```
pablooliveira, line 294: I believe this code can also be simplified as described for the double case; by removing the special treatment of the noise bit.

To avoid having to deal with the two uint64_t, I would use the native GCC __int128 type.

pablooliveira (changes requested):

pablooliveira:
I just tested this integer implementation on some basic validity tests for MCA (#299).

I'm sorry, but this implementation is buggy. Nevertheless, using integers for MCA should be sound; I have a small Python prototype working that passes my conformity tests. I will try to fix this.

pablooliveira:
Closing this, work is continued in #300.

Issue #231, opened 2021-04-06
relicensing Verificarlo


titeup:
Dear contributors,  @verificarlo/devs 
I just sent an email around to propose the relicensing of Verificarlo (except the mpfr backend).
This discussion started in https://github.com/verificarlo/fuzzy/issues/14
Here is the text of the email:
Dear Verificarlo contributors,

As discussed in PR Verificarlo/Fuzzy #14, we are discussing the idea of relicensing the project. 
Today we are using GPLv3. The historical reason is that we started in 2015 the Verificarlo project around the MPFR backend implementation of M. Frechtling. 
It turns out that Verificarlo expanded in many directions and that this backend is not central or even required in the tool anymore. 
Furthermore, for integration with other tools (such as upstreaming to LLVM or for the Fuzzy docker image) GPL could be limiting. 

After looking around and considering different options we are submitting the following action to your approval:
     1/ Release the original MCA MPFR backend in a separate repository
     2/ Relicense the rest of the code using the LLVM license based on Apache 2.0:    https://foundation.llvm.org/relicensing/LICENSE.txt

To do this relicensing, we need:
     1/ The approval of all verificarlo contributors
     2/ Parts related to missing approval must be dropped or rewriten.    

We enclosed the LLVM license file. We propose to leave one month, until  7th of May 2021, to gather consent for part 1. 
Then depending on our success, evaluate part 2 and schedule our software effort to reach a release with a new license. 
If you consent to a relicense, please answer to this PR/email with the following sentence: 
"I approve the Verificarlo relicensing under the LLVM license based on Apache 2.0 (https://foundation.llvm.org/relicensing/LICENSE.txt)"

Feel free to ask question or share your opinion on the PR. (emails is using BCC because some of the contributors did not share their email address publicly) 
We want this process to be smooth and transparent for everyone. If some of us share different opinions, please respect all of them and adopt inclusive, positive, and constructive behavior.

Thank you all,

Eric and Pablo

titeup:
I approve the Verificarlo relicensing under the LLVM license based on Apache 2.0 (https://foundation.llvm.org/relicensing/LICENSE.txt)

gkiar:
I approve the Verificarlo relicensing under the LLVM license based on Apache 2.0 (https://foundation.llvm.org/relicensing/LICENSE.txt)

mistoan:
[...]

PR #288, opened 2021-12-15
relicensing to LLVM License

titeup:
Related to issue #231 

Hi all,
I finally got everything to start relicensing. I changed all c, cpp, py etc. header file
I remove the mpfr backend since it was using external GPL code, and modify the build,  doc and tests accordingly
Change the readme and copying file
I check with some grep that we do not have keyword from the GPL license, hopefully I did not miss a spot.
I rebuild everything from a fresh clone and it still build and execute installcheck flawlessly, except the new RNG test, which was failing on my machine before the modification.

It remains:
- to discuss what to do with sigdigits submodule (work in progress with @yohanchatelain and @pablooliveira )
- checks DD license compatibility
- Changelog/release with new license
- Find why test_RNG is failing on some systems 
- and any (related) thing that can come to your mind!

Thanks,


titeup:
> * Find why test_RNG is failing on some systems

Since the test are passing on github, it seems to be a similar issue to the one experienced by @fmgjcoppens already discussed with @mistoan. I will investigate and eventually open an issue.

yohanchatelain:
> > * Find why test_RNG is failing on some systems
> 
> Since the test are passing on github, it seems to be a similar issue to the one experienced by @fmgjcoppens already discussed with @mistoan. I will investigate and eventually open an issue.

I have the same issue on my laptop.

pablooliveira:
@titeup  Regarding the test_RNG issue. I though @mistoan had fixed it after @fmgjcoppens report. Apparently not.
From what I remember from the discussion:
 - the test correct output depends on the number of OMP threads. I find this behavior annoying, maybe we should try reworking the test so it behaves deterministically. Maybe setting a fixed OMP_NUM_THREADS=... to a fixed value solves the issue ?

pablooliveira:
For DD, Verrou has relicensed to LGPL. I believe that LGPL should be compatible; but to be safe we should keep the appropriate license header on DD. We should therefore replace the old DD header by the new DD header from Verrou repository (https://github.com/edf-hpc/verrou/commit/014c42aac14773a45b8d6eb9a5bb3925b8068041).

pablooliveira:
I agree with the remaining points. 

Changelog and new release are easy.

If you need help with this, you can ping me.

pablooliveira (approved):

mistoan:
> @titeup Regarding the test_RNG issue. I though @mistoan had fixed it after @fmgjcoppens report. Apparently not. From what I remember from the discussion:
> 
>     * the test correct output depends on the number of OMP threads. I find this behavior annoying, maybe we should try reworking the test so it behaves deterministically. Maybe setting  OMP_NUM_THREADS=... to a fixed value solves the issue ?

There wasn't any actual issue opened, so it must have slipped past my radar.
I'll take note.

The solution is to fix the number of OMP threads.

pablooliveira:
> The solution is to fix the number of OMP threads.
Indeed. Now the number of tests is fixed (cf. c134129). For completeness, do you know why we need a fixed number of threads to be successful on some systems ?

mistoan:
> > The solution is to fix the number of OMP threads.
> 
> Indeed. Now the number of tests is fixed (cf. [c134129](https://github.com/verificarlo/verificarlo/commit/c134129e40e469296a784363e1f258f563f45845)). For completeness, do you know why we need a fixed number of threads to be successful on some systems ?

I couldn't say for sure, but for me, the most probable cause would be the fact that they were running these tests in virtual machines, so something was probably going on when the cores were being allocated, between the runs.

yohanchatelain:
Could it be due to a type mismatch? because `lrand48_r` is expecting `int64_t` but we use `uint64_t`.
https://github.com/verificarlo/verificarlo/blob/f4b554b833dca4284803f6ada72b2e1cf2465756/src/common/options.c#L96-L104

titeup:
I do not think it has any influence, lrand expect a pointer on 64bit allocated memory.
I suppose Mattei used uint because it just need a random bit field with at least 53 random bits.
"The lrand48() and nrand48() functions return non-negative long integers uniformly distributed between 0 and 2^31."
So he needs two of it (l99, l100), shift and add l101). and then he can return 64 random bits. Nonetheless, I wonder if this is a valid way to simply concatenate like that. 
Regarding the threading, I suppose it is worth checking that the number of random state structure is matching the number of threads (printing in the log during the test and |uniq )

Note: this pull request is closed and its pointing to relicensing (I plaid guilty to have introduced this side channel), maybe we should reopen @mistoan issue/PR on reentrant number generator and continue the discussion here).

pablooliveira:
> Note: this pull request is closed and its pointing to relicensing (I plaid guilty to have introduced this side channel), maybe we should reopen @mistoan issue/PR on reentrant number generator and continue the discussion here).

I agree. Maybe we can continue this as a [discussion](https://github.com/verificarlo/verificarlo/discussions) and if needed open an issue, when confirmed).

titeup:
https://github.com/verificarlo/verificarlo/discussions/296

PR #289, opened 2021-12-15
Improve test suite

yohanchatelain:
- Fix issue with wrong python3 path
	- Parallelize some tests
	- Move compilation in outer loop
	- Decrease loop size in test_sparsity


pablooliveira:
This looks pretty cool, thanks !

pablooliveira:
There is a new depency `parallel`. I'm not against it. But the dependencies of verificarlo have been growing a lot lastly. This worries me because it raises the adoption cost for new users.

Since some dependencies are only required by tests, such as parallel or bigfloat. Maybe we should separate `test` dependencies from `package` dependencies ?

yohanchatelain:
Are you worried about the number of dependencies or the fact that it can discourage the new user?
For the latter, we could write a script to download dependencies automatically, something like `python3 install.py dependencies <--full> <--test> <--minimal>`. 
What do you think?

pablooliveira:
I'm not sure about automatically installing things since the user might want to install system wide or in a local PYTHONPATH and this can cause problems with the path also.

But in any case the problem of dependencies and this issue are separate. I'm ok with merging this issue as it is.
Just wanted to start discussing how we deal with dependencies because I see a growing problem :-)

pablooliveira (approved):

PR #290, opened 2021-12-15
Redirect logger_info to stdout instead of stderr

yohanchatelain:
- Redirect logger_info to stdout instead of stderr since it is not an error output
-  Modify VFC_BACKENDS_LOGGER=False behaviour to disable logger_info outputs.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,5 +1,10 @@
 Verificarlo CHANGELOG
 
+# [Unreleased]
+
+## Changed
+  * VFC_BACKENDS_LOGGER=False behaviour. Now truly disable logger_info output.
+  * Redirect logger_info to stdout instead of stderr
 
 # [v0.7.0] 2022/01/22
 
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -2,6 +2,8 @@ Verificarlo CHANGELOG
 
 # [Unreleased]
 
+## Added
+  * Add VFC_BACKENDS_LOGFILE environment variable to redirect logger_info to a file
 ## Changed
   * VFC_BACKENDS_LOGGER=False behaviour. Now truly disable logger_info output.
   * Redirect logger_info to stdout instead of stderr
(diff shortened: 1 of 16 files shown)


pablooliveira:
Hi @yohanchatelain. Thanks for this PR. 
One advantage I see when keeping verificarlo logger messages in stderr is that we can separate the stdout from the original program from the verificarlo messages.

I also agree that info messages are not errors and would be better in stdout.

Therefore, I'm in two minds about this issue. Could you please explain why this change is important for you ? Thanks !

yohanchatelain:
Hi @pablooliveira, I got your point. The issue was that we ran a pipeline instrumented with verificarlo that launches several third applications and checks if the execution of each app was correct. In particular, it checks that the `stderr` output was empty, and since verificarlo outputs on this stream, it raises an error. Since `VFC_BACKENDS_LOGGER=False` does not really suppress the outputs, we had no solutions to prevent this issue. 
What about adding an environment variable to specify the file descriptor to which redirect logger output?
Then it could be used like this:
`VFC_BACKENDS_LOGGER_FD=stderr ./a.out 2> log` to redirect on `stderr`.
And we can let it on `stdout` by default.
What do you think?

pablooliveira:
I see, thanks for explaining the rationale.  A few questions:
  - Why does VFC_BACKENDS_LOGGER=False not suppress the outputs ?
  - We could instead have a VFC_BAKCKENDS_LOGFILE="verificarlo.log" that redirects to a file ? I think it will be more useful to the user than LOGGER_FD.

yohanchatelain:
> I see, thanks for explaining the rationale. A few questions:
> 
> * Why does VFC_BACKENDS_LOGGER=False not suppress the outputs ?
https://github.com/verificarlo/verificarlo/blob/ac69629681c6b6c78101c2c6abd3274f36f07306/src/common/logger.c#L129-L142
The guard `logger_enabled` does not include the `vprintf`. Originally, the `printf` was included for `logger_{info,warning,error}` but by disabling the logger, we were disabling the error as well. 
> * We could instead have a VFC_BAKCKENDS_LOGFILE="verificarlo.log" that redirects to a file ? I think it will be more useful to the user than LOGGER_FD.
Sure, makes more sense :+1:

pablooliveira (approved):
Thanks, looks good !

PR #298, opened 2022-02-17
Add `interflop_call` function into `interflop_backend_interface_t`

yohanchatelain:
- Allows the user to call low-level backend functions directly
- Only implemented in the libinterflop_mca.so backend
- INTERFLOP_CALL_INEXACT: allows one to perturb one floating-point value directly. Perturbation is done at VIRTUAL_PRECISION - 1 to always introduce a noise, even when t=24 or t=53.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -4,6 +4,10 @@ Verificarlo CHANGELOG
 
 ## Added
   * Add VFC_BACKENDS_LOGFILE environment variable to redirect logger_info to a file
+  * Add `interflop_call` function into interflop_backend_interface_t interface.
+  Allows user to directly call low-level backend functions
+  * Add INTERFLOP_CALL_INEXACT allowing to directly perturb one fp value through `interflop_call` mechanism
+  * Add test_usercall_inexact to test the INTERFLOP_CALL_INEXACT user call
 ## Changed
   * VFC_BACKENDS_LOGGER=False behaviour. Now truly disable logger_info output.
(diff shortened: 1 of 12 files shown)


pablooliveira:
Hi @yohanchatelain, I see you are actively working on this. Ping me when you are finished if you need a reviewer. Thanks for the useful feature.

yohanchatelain:
> Hi @yohanchatelain, I see you are actively working on this. Ping me when you are finished if you need a reviewer. Thanks for the useful feature.

Hi @pablooliveira, thanks for your help. The PR is ready now, even if I did not document this feature in the doc. I would like to discuss the API before documenting it.

pablooliveira:
I like the interface, the way you programmed it seems to be general enough to add a variety of `interflop_calls` depending on user needs.

yohanchatelain:
> I like the interface, the way you programmed it seems to be general enough to add a variety of `interflop_calls` depending on user needs.

Nice. I reused from memories the one designed for the interflop project

yohanchatelain:
Hi @pablooliveira, I added the documentation. After merging the PR, it could be nice to make a new release version. What do you think?

tcheulingdemeudon (approved):

pablooliveira:
Regarding the new release; could we wait a little ? I'm still checking the new mca-int backend. In particular, I think we should disallow using it for t < p since the result is biased.

pablooliveira:
Thanks !

PR #299, opened 2022-02-22
Add test to check MCA SR properties

pablooliveira:
Since removing the MPFR reference backend we are missing good validity
tests for MCA backends.

This test checks basic statistical properties of MCA RR mode with
t=53 and t=24.

Test logic is detailed in check.py

### DIFF
--- a/tests/test_sr/check.py
+++ b/tests/test_sr/check.py
@@ -18,34 +18,42 @@ def read(name):
         freqs = freq(data)
     return freqs
 
+# TODO: Determine the tolerance precisely with a confidence interval depending
+# on the sampled iterations.
 TOLERANCE=15
 
 if __name__ == "__main__":
+    import sys
+    assert(len(sys.argv) == 2)
+    its = int(sys.argv[1])
     f = read("binary32-23")
     assert(len(f) == 1)
-    assert(f['0x1.4000020000000p+0'] == 100)
-    # For inexact24(1.0 + 2**-23) since the computed value is exact in 24 bits,
+    assert(f['0x1.4000020000000p+0'] == its)
+    # For inexact24(1.25 + 2**-23) since the computed value is exact in 24 bits,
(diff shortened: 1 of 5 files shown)


tests/test_sr/check.py
```diff
@@ -0,0 +1,68 @@
...
+
+
+def read(name):
+    """ read filename and returns the distribution of outputs as a dictionnary """
+    with open(name, "r") as f:
+        data = [x.strip() for x in f.readlines()]
+        freqs = freq(data)
+    return freqs
+
+TOLERANCE=15
```
titeup, line 21: How did you define Tolerance? I suppose it is related to the number of samples with a given confidence, maybe we can justify it in comments. (if I am right the number of 'samples' is the number of iteration of the loop in the C code (100)) 
We a single addition we can assume normal distribution and apply the confidence interval for it?

tests/test_sr/sr-binary32.c
```diff
@@ -0,0 +1,16 @@
...
+#include <assert.h>
+#include <stdlib.h>
+
+int main(int argc, char * argv[]){
+    assert(argc == 2);
+    int p = atoi(argv[1]);
+    float a = 1.25;
+    float b = pow(2, -p);
+
+    for (int i=0; i < 100; i++) {
```
titeup, line 12: can we make the 100 a param and relates it to the tolerance define in the py script, both of them being determine in the run script at the same place?

titeup (approved):
It looks valide to me as is. However I submit you a comment about tolerance and number of samples that could be an improvement.

tests/test_sr/sr-binary32.c
pablooliveira: Ok.

tests/test_sr/check.py
pablooliveira: I added a comment. Thanks.

PR #300, opened 2022-02-22
Mca integer backend

pablooliveira:
This pull request brings a version of the MCA back-end where the operations regarding the noise are done using integer and bitwise operations.

Continue work started in #287.


pablooliveira:
In commit 40936b5, I have reduced the random pool used.
For a binary32 number, I use 32 bits of noise.
For a binary64 number, I use 64 bits of noise.

Since most of noise is right-shifted (since we usually work at t=p) the low bits of the noise are in any case lost most of the time when working in precision 2.p.

pablooliveira:
Requesting review / feedback from @titeup @mistoan 

I have continued the work started by @mistoan for implementing MCA integer backed.

- Both binary64 and quad noise routines have been  fixed. Now they pass the `test_sr` test. Among multiple problems, there was an issue with the computation of the shift to find the correct noise magnitude.

- Random routines in `src/common/` have been fixed. Since they did only return 63 or 31 bits (leaving the signed bit blank). This simplifies a lot the code, which does not need to do specific things with the sign bit.

- quad routine has been simplified quite a bit by using the __int128 type. This way I do not have to manage explictly the carry between the low and high part of the mantissa.

- Finally I have reduced the amount of random bits used. My rationale is in the above comment.

pablooliveira:
Tested successfully on Kahan -O0 vs -O3; comparable histograms with mca and mca-int backends.

pablooliveira:
Performance tests done on NAS CG (Class A), setup available in tests/test_fortran_NAS/NPB3.0:
Test machine: Intel(R) Core(TM) i7-9850H CPU @ 2.60GHz

- Non instrumented (flang-7): 0.80s 
- libinterflop_ieee.so: 6.41s (x 8 slowdown)
- libinterflop_mca.so: 173.68s (x 217 slowdown)
- libinterflop_mca_int.so: 128.29s (x 160 slowdown)

Speedup of mca_int over mca = x 1.35 (double precision)

Next will do a comparison between double (using 128) against single (using binary64)

titeup:
> Performance tests done on NAS CG (Class A), setup available in tests/test_fortran_NAS/NPB3.0: Test machine: Intel(R) Core(TM) i7-9850H CPU @ 2.60GHz
> 
> * Non instrumented (flang-7): 0.80s
> * libinterflop_ieee.so: 6.41s (x 8 slowdown)
> * libinterflop_mca.so: 173.68s (x 217 slowdown)
> * libinterflop_mca_int.so: 128.29s (x 160 slowdown)
> 
> Speedup of mca_int over mca = x 1.35 (double precision)
> 
> Next will do a comparison between double (using 128) against single (using binary64)

Did you capture the time spend in each fct? I wonder about the random generator overhead since it changed recently (to use reentrant one)

pablooliveira:
> Did you capture the time spend in each fct? I wonder about the random generator overhead since it changed recently (to use reentrant one)

I did not. But both mca backends are using the same library rand48. But different generating functions.

pablooliveira:
Some more performance numbers. This time on kahan compiled with -O0 (to measure simple instruction overhead). Test is available in `tests/test_kahan`.
On 100 million array elements.

= Binary32 =

- Non instrumented (clang-7), 1.34 s
- libinterflop_ieee.so 2.36s
- libinterflop_mca.so, 6.28s
- libinterflop_mca_int.so, 7.76s

= Binary64 =

- Non instrumented (clang-7), 1.34 s
- libinterflop_ieee.so 2.34s
- libinterflop_mca.so, 105s
- libinterflop_mca_int.so, 64s

So on simple precision, the integer backend seems to be a tad slower.
But it provides large benefits on double precision.

I did not try to optimize the backend much, or play with compiler flags on the backend. It's possible we can be a bit faster.

titeup (approved):
All clear!

titeup:
> All clear!

PASS: testplan.test 31 ./test_sr

PR #303, opened 2022-06-22
badges update in README

titeup:
Hi, I had these small commit pending after we relicence and forgot to do the pull request.
They are still worth considering but I will not object if they are discarded.

Thanks,

Eric


pablooliveira (approved):

pablooliveira:
Thanks @titeup, looks good, merged !

PR #304, opened 2022-06-23
Fix options in MCA integer backend

pablooliveira:
Some of the user options such as custom precisions and relative error are currently unsupported in the MCA integer backend.

This PR removes the options explicitly and document the missing options.

### DIFF
--- a/src/backends/interflop-mca-int/interflop_mca_int.c
+++ b/src/backends/interflop-mca-int/interflop_mca_int.c
@@ -365,16 +365,8 @@ _INTERFLOP_OP_CALL(double, mul, mca_mul, _mca_binary64_binary_op)
 _INTERFLOP_OP_CALL(double, div, mca_div, _mca_binary64_binary_op)
 
 static struct argp_option options[] = {
-    {key_prec_b32_str, KEY_PREC_B32, "PRECISION", 0,
-     "select precision for binary32 (PRECISION > 0)", 0},
-    {key_prec_b64_str, KEY_PREC_B64, "PRECISION", 0,
-     "select precision for binary64 (PRECISION > 0)", 0},
     {key_mode_str, KEY_MODE, "MODE", 0,
      "select MCA mode among {ieee, mca, pb, rr}", 0},
-    {key_err_mode_str, KEY_ERR_MODE, "ERROR_MODE", 0,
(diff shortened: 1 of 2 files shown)


pablooliveira:
Hi @titeup. As previously discussed MCA int does not properly support custom (the result will be different than what MCA quad precisions nor does it support absolute error mode.

This PR forbids these modes from the user interface.

src/backends/interflop-mca-int/interflop_mca_int.c
```diff
@@ -393,22 +388,20 @@ error_t parse_opt(int key, char *arg, struct argp_state *state) {
     /* precision for binary32 */
     errno = 0;
     val = strtol(arg, &endptr, 10);
-    if (errno != 0 || val <= 0) {
-      logger_error("--%s invalid value provided, must be a positive integer",
+    if (errno != 0 || val != MCA_PRECISION_BINARY32_DEFAULT) {
+      logger_error("--%s invalid value provided, MCA integer does not support "
+                   "custom precisions",
                    key_prec_b32_str);
-    } else {
```
titeup, line 399: So you remove the opportunity for the user to use virtual prec. But your are also forbiding it for an expert user that could live with some bias under certain circumstance?  Can't we have an hidden parameter and keep the code?

pablooliveira: I think it's better to forbid it in MCA int. The result distribution is biased when using a precision different than the working one; and the operation is different than what one would expect for MCA RR.

 If an expert user still wants to play with it, it's pretty trivial to switch it back, don't you think ? Better to add it back then, than keep the complexity of a hidden option ?

pablooliveira: Btw, I keep the option in MCA quad. So it's still possible to have virtual precision in MCA (just with a different backend).

titeup: It is true that we never used so much virtual precision in mca since we have vprec for the precision work. In  the reare case we would like both I suppose mca quad is fast enough, and if not then we might pay the price of having this hidden option (like --force or something to force it to be accepted).

titeup:
@pablooliveira I leave it to you to merge as soon as you are ready.

PR #308, opened 2022-07-08
MCA int optimizations

yohanchatelain:
- Merge operations to save cycles
- Move to xoshiro RNG that is faster than jrand48


pablooliveira:
This looks really cool :)
Happy to review once you stabilize the PR !

src/backends/interflop-mca-int/interflop_mca_int.c
```diff
@@ -165,30 +166,23 @@ static __thread rng_state_t rng_state;
...
-  int64_t x_s64 = *(int64_t *)(x);
+  binary64 *b64 = (binary64 *)x;
 
   // amount by which to shift the noise term sign (1) + exp (11) + noise
   // exponent
-  shift = 1 + DOUBLE_EXP_SIZE - exp;
+  const uint32_t shift = 1 + DOUBLE_EXP_SIZE - exp;
 
   // noise is a signed integer so the noise is centered around 0
   // only 32 bits of noise are used, they are left aligned in a signed 64 bit
```
pablooliveira, line 177: What is the advantage of using 64 bits of noise (get_rand_uint64) here over 32 ? 
I'm fine with both but would love the explanation. In any case, if you keep 64 random bits could you please update comment in line 177 ?

src/backends/interflop-mca-int/interflop_mca_int.c
```diff
@@ -165,30 +166,23 @@ static __thread rng_state_t rng_state;
...
-  noise = ((uint64_t)_get_rand_uint32(rng_state, &global_tid_lock, &global_tid)
-           << 32);
+  int64_t noise = get_rand_uint64(rng_state, &global_tid) & 0xFFFFFFFF00000000;
```
pablooliveira, line 178: Remove the mask here. Better to use the 32 extra random bits since we paid for them :-)

The code before was just a trick to avoid using 64 random bits; using 32 gave similar results and was cheaper.

src/backends/interflop-mca-int/interflop_mca_int.c
```diff
@@ -201,28 +195,22 @@ static void _noise_binary128(__float128 *x, const int exp,
...
   // and sign bit will be extended
-  noise = noise >> shift;
+  noise.i128 >>= shift;
 
   // Add the noise
-  x_s128 += noise;
-
-  // Convert back to __float128
-  *x = *(__float128 *)(&x_s128);
+  b128->i128 += noise.i128;
```
pablooliveira, line 213: Are you sure the masks are not more costly ? I find the previous cast pretty clean here. But if you prefer the unions, I'm ok.

tests/test_daz_ftz/test.sh
```diff
@@ -37,7 +37,7 @@ export VFC_BACKENDS_SILENT_LOAD="TRUE"
 rm -f run_parallel
 for REALTYPE in "float" "double"; do
     compile $REALTYPE
-    for BACKEND in "libinterflop_mca.so" "libinterflop_bitmask.so"; do
+    for BACKEND in "libinterflop_mca.so"; do # "libinterflop_bitmask.so"; do
```
pablooliveira, line 40: Why do you remove bitmask here ?

pablooliveira (changes requested):
This looks really cool. I gave some feedback; nothing major, mostly questions.

src/backends/interflop-mca-int/interflop_mca_int.c
yohanchatelain: Yes it is less costly and it's actually where we save cycles.
With the last cast, the compiler generates `mov` instructions.
image

Using the union, we generate `add` instructions, which is faster.
image

tests/test_daz_ftz/test.sh
yohanchatelain: It was a mistake, thanks for pointing it out!

src/backends/interflop-mca-int/interflop_mca_int.c
pablooliveira: Oh wow, I would have never guessed :clap:

src/backends/interflop-mca-int/interflop_mca_int.c
yohanchatelain: Since the RNG only generates 64 bits at once, it's the same cost to ask for 64 or 32 bits. I added the function to generate 32 random bits for convenience only.
I'll fix the comment. Thanks!

yohanchatelain:
I removed the comment in the commit about OneAPI DPC because it's not related to the purpose of the PR.
Moreover, we should add more tests to be sure that nothing is broken.

pablooliveira:
Ok thanks @yohanchatelain. Do you want me to approve the PR now or wait until you are finished with the tests ?

titeup:
> I removed the comment in the commit about OneAPI DPC because it's not related to the purpose of the PR. Moreover, we should add more tests to be sure that nothing is broken.

For the record, we discuss this offline, but the original meaning of your ref to Icx/dpc was that you get better perf while compiling the backend with icx. This could be still relevant to offer it as an option/document if it makes a significant difference on some platform.

src/backends/interflop-bitmask/Makefile.am
```diff
@@ -1,9 +1,9 @@
 lib_LTLIBRARIES = libinterflop_bitmask.la
 libinterflop_bitmask_la_SOURCES = interflop_bitmask.c ../../common/logger.c ../../common/options.c
-libinterflop_bitmask_la_CFLAGS = -DBACKEND_HEADER="interflop_bitmask"
+libinterflop_bitmask_la_CFLAGS = -DBACKEND_HEADER="interflop_bitmask" -O3
```
titeup, line 3: did you check O3 with various compiler? it can make a huge diff from one to another

src/backends/interflop-bitmask/Makefile.am
```diff
@@ -1,9 +1,9 @@
...
 if WALL_CFLAGS
 libinterflop_bitmask_la_CFLAGS += -Wall -Wextra
 endif
 libinterflop_bitmask_la_LDFLAGS = -lm
-libinterflop_bitmask_la_LIBADD = 
+libinterflop_bitmask_la_LIBADD = ../../common/rng/libvfc_rng.la
```
titeup, line 8: As a general comment, nothing to be changed yet, but to discuss for later. shall we keep relative path? Or use verificarlo_home variable?

src/backends/interflop-ieee/Makefile.am
```diff
@@ -1,6 +1,6 @@
 lib_LTLIBRARIES = libinterflop_ieee.la
 libinterflop_ieee_la_SOURCES = interflop_ieee.c ../../common/printf_specifier.c ../../common/logger.c
-libinterflop_ieee_la_CFLAGS = -DBACKEND_HEADER="interflop_ieee"
+libinterflop_ieee_la_CFLAGS = -DBACKEND_HEADER="interflop_ieee" -O3
```
titeup, line 3: same as before, just to remeber there is more than one

src/backends/interflop-cancellation/interflop_cancellation.c
```diff
@@ -101,7 +98,7 @@ static inline double _noise_binary64(const int exp, rng_state_t *rng_state) {
       _init_rng_state_struct(&RNG_STATE, ((t_context *)CTX)->choose_seed,      \
                              (unsigned long long)(((t_context *)CTX)->seed),   \
                              false);                                           \
-      *Z = *Z + _noise_binary64(e_n, &RNG_STATE);                              \
+      *Z += _noise_binary64(e_n, &RNG_STATE);                                  \
```
titeup, line 101: I am ok since it is functionally eq, but not like it much when doing operations beyond simple integer induction variables.

src/common/logger.c
```diff
@@ -128,19 +130,13 @@ bool is_logger_colored(void) {
   }
 }
 
-pid_t get_tid() {
-#pragma GCC diagnostic push
-#pragma GCC diagnostic ignored "-Wimplicit-function-declaration"
-  return gettid();
-#pragma GCC diagnostic pop
-}
+pid_t get_tid() { return syscall(__NR_gettid); }
```
titeup, line 133: why not using gettid() directly instead of redoing a wrapper which does the same.(for older libc?) (same question from the previous version of the wrapper)

src/common/rng/xoroshiro128.c
```diff
@@ -0,0 +1,70 @@
...
+
+/*********************** XOROSHIRO128++ FUNCTIONS ****************************\
+ *                                                                           *\
+ *  Written in 2019 by David Blackman and Sebastiano Vigna (vigna@acm.org)   *\
+ *                                                                           *\
+ *  To the extent possible under law, the author has dedicated all copyright *\
+ *  and related and neighboring rights to this software to the public domain *\
+ *  worldwide. This software is distributed without any warranty.            *\
+ *                                                                           *\
+ *  See <http://creativecommons.org/publicdomain/zero/1.0/>.                 *\
```
titeup, line 28: is this license compatible with apache2.0? I suppose yes, but we need to be 100% positive.

tests/test_cancellation/result.txt
```diff
@@ -1,231 +1,231 @@
 Info [interflop_cancellation]: interflop_cancellation: loaded backend with tolerance = 0
```
titeup, line 1: is this file suppose to be comited on the git?

titeup (changes requested):
not necessarily expecting changes everywhere, but answers in the various comments

src/backends/interflop-bitmask/Makefile.am
titeup: yes, icx, clang, and gcc.
And even test if o3 vs O0 and O2 for clang.
 I kind of remember long ago with the first version of quad, O3 was not working with a trick I brrowed from cadna to do up or down Stochastic rounding (1/2-1/2 probability), and we had this very same discussion for the other backends and conclude we do not have enough test to be confident at the time. 
I am more confident now about the test, but I rise the flag, maybe we should discuss it further with you and Pablo about potential impact and if this should be the default behavior or optional to use O3.

yohanchatelain: I made some tests on NAS BT S and `clang` is faster than `gcc`.
| Backends version | Compiler  | IEEE |  mca int | mca |  
|------------|--------------|---------|--------|-----------|
| Original | `gcc-9`     |  1.22s         |   38.03s      | 58.94s  |
| Original | `clang-9`  |   1.25s        |    31.25s     | 54.11s |
| New | `gcc-9`     |   1.22s        |   30.35s      |    56.78s         |
| New | `clang-9`  |   1.26s        |    18.15s | 45.22s    |

src/backends/interflop-ieee/Makefile.am
yohanchatelain: There is no real improvements for the IEEE backend.

src/backends/interflop-cancellation/interflop_cancellation.c
yohanchatelain: As you like, it's a matter of taste

src/common/logger.c
yohanchatelain: Yes the main reason is that `gettid` is not available with older version of linux kernel.

src/backends/interflop-bitmask/Makefile.am
yohanchatelain: I opened an issue to discuss it later but it would make the Makefile cleaner. :+1:

src/backends/interflop-bitmask/Makefile.am
titeup: I was more worried about correctness, but it is very good to have some perf figure. It makes a huge difference, enough to spend the time looking for correctness if there is a glitch somewhere ;)

src/common/rng/xoroshiro128.c
yohanchatelain: I checked and normally it is the most [permissive license](https://creativecommons.org/publicdomain/zero/1.0/)

> The person who associated a work with this deed has dedicated the work to the public domain by waiving all of his or her rights to the work worldwide under copyright law, including all related and neighboring rights, to the extent allowed by law.
> 
> You can copy, modify, distribute and perform the work, even for commercial purposes, all without asking permission. See Other Information below.

tests/test_cancellation/result.txt
yohanchatelain: Yes, it's the reference to compare against.

titeup: So why is there some changes in all the ref values despite the benchmark code itself has not change?

src/backends/interflop-bitmask/Makefile.am
yohanchatelain: > yes, icx, clang, and gcc. And even test if o3 vs O0 and O2 for clang. I kind of remember long ago with the first version of quad, O3 was not working with a trick I brrowed from cadna to do up or down Stochastic rounding (1/2-1/2 probability), and we had this very same discussion for the other backends and conclude we do not have enough test to be confident at the time. I am more confident now about the test, but I rise the flag, maybe we should discuss it further with you and Pablo about potential impact and if this should be the default behavior or optional to use O3.

For mca quad, `-O3` breaks the tests so it's definitely too aggressive, but maybe with specific flags (like `-ffp-model=strict` for `clang`) it could work but I did not test it. We could also add `-march=native` to trigger ISA but it will trouble docker images shipping, as the arch is not the same.

src/common/rng/xoroshiro128.c
titeup: It seems like CC in Apache2.0 program is doable but not the opposite. It seems to be more complicated than expected matter. I will try to find a qualified person to give a definitive answer, especially if we can sublicence, if we need to retain header, license file for that part and so on..
You enclosed the CC paragraph in a file starting with our apache2.0 header, and did not put a marker when the CC code end. Maybe the correct solution is to remove the apache header, move up the CC header at the begining of the file and have this file being CC... 
@pablooliveira any opinion/advice on your side?

pablooliveira: This is CC0  so equivalent to a Public domain license. Should be fine to include it. You can keep the original header.

tests/test_cancellation/result.txt
pablooliveira: I believe because the test depends on the random seed. Matei had the same issue when he was working on the new RNG interface.

src/common/logger.c
pablooliveira: Syscall is fine and should be compatible with both old and new versions.

yohanchatelain: Yes, it's the most compatible one. I tested by checking the kernel version but it was not stable across compilers so I went back on `syscall`

tests/test_cancellation/result.txt
yohanchatelain: Exactly

> I believe because the test depends on the random seed. Matei had the same issue when he was working on the new RNG interface.

Exactly.

src/common/rng/xoroshiro128.c
titeup: @pablooliveira to be sure we are on the same page, you also suggest to remove our apache header from this file and keep only the CC header?

pablooliveira: @titeup yes, that seems good for me. Keep only the original CC0 header on that file.

pablooliveira (approved):

pablooliveira:
I think all the changes requested by @titeup have been addressed. Merging this !

PR #310, opened 2022-09-05
Adding usercall to change precision and range at runtime for vprec and mca backends

vincentAlba:
Add user calls INTERFLOP_SET_PRECISION_BINARY32 and INTERFLOP_SET_PRECISION_BINARY64 to change precision on MCA and VPREC backends

Add user calls INTERFLOP_SET_RANGE_BINARY32 and INTERFLOP_SET_PRECISION_BINARY64 to change range in the VPREC backends

Also add a test "test_usercall_change_precision" to check if those user calls works correctly


pablooliveira (approved):
This PR looks good ! Maybe we should document the feature in the user manual, but besides that it can be merged.

PR #313, opened 2022-09-23
Fix CI workflow templates by setting safe.directory

PurplePachyderm:
After a recent security vulnerability, Git changed the way it looks for repositories in parent directories (https://github.blog/2022-04-12-git-security-vulnerability-announced/). The easiest fix seems to set the safe.directory variable at the beginning at the workflow to set an exception to the old behavious, which caused the repository to not be detected in the CI runner.


pablooliveira (approved):
LGTM !

PR #317, opened 2022-10-04
Improve test to plot the resulting histogram

pablooliveira:
Add a plot script to `test_linear_system_kahan`


titeup (approved):
All good to me.
I suppose including this plot script is for teaching/demo purpose? In which case why not making a separate test.sh (eg test_plot.sh) to not interact with the production of the result for the 'real' test?

pablooliveira:
I messed up updating the branch, sorry. Will open again.

PR #319, opened 2022-10-04
Improve test

pablooliveira:
Followup of #317 which I closed by mistake when updating the branch with the requested change.

### DIFF
--- a/tests/test_linear_system_kahan/clean.sh
+++ b/tests/test_linear_system_kahan/clean.sh
@@ -1,3 +1,3 @@
 #!/bin/bash
 
-rm -Rf *~ output test test.log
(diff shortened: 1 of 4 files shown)


titeup (approved):
All good to me!
And no additional comments ;)

titeup:
In the last version using test.sh I realize that maybe clean.sh was not modified accordingly, It is not necessary now, or is it? If someone play with the test, later do a complete clean of the project would we like to make sure that we remove his x0 trace file, or not?

pablooliveira:
If only test.sh is ran, then clean.sh will still work since x0.dat is created in test_plot.sh. But if you prefer, I can add x0.dat to clean.sh also ?

Issue #333, opened 2023-02-06
Support LLVM 15.0.6 (Ubuntu 22.04 tests)


pjaaskel:
I tested with Ubuntu 22.04 and Clang 15.0.6. It built quite easily, but required patching configure.ac to accept 15.0.6 and installing bigfloat package from git main since the 0.4.0 didn't build via pip3 (see https://github.com/mdickinson/bigfloat/issues/115).

There are some errors in make installcheck though:
```
PASS: testplan.test 1 ./test_absolute_error
FAIL: testplan.test 2 ./test_bitmask_backend
PASS: testplan.test 3 ./test_branch_instrumentation
PASS: testplan.test 4 ./test_cancellation
PASS: testplan.test 5 ./test_cpp
PASS: testplan.test 6 ./test_cython
PASS: testplan.test 7 ./test_daz_ftz
FAIL: testplan.test 8 ./test_ddebug_archimedes
FAIL: testplan.test 9 ./test_ddebug_canc
FAIL: testplan.test 10 ./test_ddebug_exclude
FAIL: testplan.test 11 ./test_ddebug_gen_include
FAIL: testplan.test 12 ./test_ddebug_integration
PASS: testplan.test 13 ./test_ffastmath
PASS: testplan.test 14 ./test_fix_seed
PASS: testplan.test 15 ./test_fortran
PASS: testplan.test 16 ./test_fortran_NAS
PASS: testplan.test 17 ./test_fortran_vfc_probes
PASS: testplan.test 18 ./test_fp80
FAIL: testplan.test 19 ./test_ieee_backend
PASS: testplan.test 20 ./test_ieee_backend_debug_binary
PASS: testplan.test 21 ./test_include_exclude
PASS: testplan.test 22 ./test_kahan
PASS: testplan.test 23 ./test_linear_system_kahan
PASS: testplan.test 24 ./test_logger
FAIL: testplan.test 25 ./test_mca_rng
PASS: testplan.test 26 ./test_per_function
PASS: testplan.test 27 ./test_rr_mode_exact
PASS: testplan.test 28 ./test_save_temps_option
PASS: testplan.test 29 ./test_shell_expansion
PASS: testplan.test 30 ./test_sparsity
PASS: testplan.test 31 ./test_sr
PASS: testplan.test 32 ./test_static_mode
PASS: testplan.test 33 ./test_tchebychev
FAIL: testplan.test 34 ./test_unstable_branches
[...]

PR #335, opened 2023-02-16
Sycl/pocl support

titeup:
Following issue https://github.com/verificarlo/verificarlo/issues/333
And discussion https://github.com/verificarlo/verificarlo/discussions/334

I apply the patch from Pekka, did not add any dependency.
I will open separate  issues for the pb on the tests with latest ubuntu since it is unrelated to pocl support.

This modification allows to compile to bytecode .bc with verificarlo. 
Local test on previous ubuntu pass even with build of LLVM16. This still need to go through CI workflow.


configure.ac
```diff
@@ -24,7 +24,7 @@ AC_MSG_RESULT([$GCC_VERSION])
 AX_COMPARE_VERSION([$GCC_VERSION],[ge],['6.0.0'],[],[AC_MSG_ERROR([At least GCC version 6 is required])])
 AC_DEFINE_UNQUOTED([GCC_PATH], ["$CC"], [GCC path (for compiling)])
 AC_SUBST(GCC_PATH, $CC)
-AX_LLVM([4.0],[15.0.1],[all])
+AX_LLVM([4.0],[15.0.6],[all])
```
pablooliveira, line 27: I think we need to revert this until we fix all the bugs in #333.

configure.ac
```diff
@@ -175,7 +182,7 @@ fi
 
 # Check for Python modules
 AX_PYTHON_MODULE(numpy, fatal)
-AX_PYTHON_MODULE(bigfloat, fatal)
+AX_PYTHON_MODULE(bigfloat)
```
pablooliveira, line 185: This seems a good idea, since `bigfloat` is only a dependency for tests.

src/libvfcfuncinstrument/libVFCFuncInstrument.cpp
```diff
@@ -391,6 +391,7 @@ struct VfclibFunc : public ModulePass {
   virtual bool runOnModule(Module &M) {
     TargetLibraryInfoWrapperPass TLIWP;
 
+    std::cerr << "vfcfuncinst" << std::endl;
```
pablooliveira, line 394:
```suggestion
```

configure.ac
pablooliveira, line 27:
```suggestion
AX_LLVM([4.0],[15.0.1],[all])
```

tests/test_ieee_backend/test.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/sh
 set -e
-verificarlo-c --save-temps -O0 test.c -o test
+verificarlo-c -ffp-contract=off --save-temps -O0 test.c -o test
```
pablooliveira, line 3: I think `-ffp-contract=off` should be added to the default verificarlo flags in the driver script.  Until we fix FMA support we need it everywhere and not only on this test.

tests/test_vprec_instrument/test.sh
```diff
@@ -1,5 +1,6 @@
 #!/bin/bash
 #set -e
```
pablooliveira, line 2:
```suggestion
```

tests/test_ieee_backend/test.sh
pablooliveira, line 3:
```suggestion
verificarlo-c --save-temps -O0 test.c -o test
```

pablooliveira (commented):
Thanks @pjaaskel and @titeup, this looks good !

A few small suggestions. The only important issue is that `-ffp-contract=off` should be always used (and not only in the test) until we fix FMA support. I will push a suggested patch to this PR.

pablooliveira:
@titeup in d4a81ad I moved the flag that disables `ffp-contract` to the driver. Could you please check that this is still ok? I could not reproduce the FMA issue on my setup. :pray:

configure.ac
titeup: same as openmp, this is why I was wondering libomp this shouldn't be added to dependency list. On the other hand as we discussed monday, a mechanism to disable the test/part of the test concerned by a given dependency might be a good idea.

configure.ac
titeup: No it works with LLVM up to 16 from the git on ubuntu 18
Beyond FMA fp-contract starting at llvm14, the remaining bug in  https://github.com/verificarlo/verificarlo/issues/333 relates to the default packages in ubunutu 22.04.
So I do not see the pb of going to 15.0.6

titeup: see previous comment

tests/test_ieee_backend/test.sh
titeup: Comment for the record, I will validate your solution.
This is what I had done first but I changed my mind. 
First because the idea was to open a separate issue for it that will be solved after @yohanchatelain PR to support FMA (i will open it anyway to remind us to remove this flag during the PR)
Second because I am not sure of other negative aspect of forcing -ffp-contract=off
What makes me think it is fair to reconsider the situation as you suggest is that even if outside the test_ieee, the fact of not instrumenting the FMA will not cause a crash, not instrumenting some of the operation might be a worse behavior than changing a bit the arithmetic with -ffp-contract....
The only right fix is to support FMA... (and reduce)

titeup: BTW I just tested your version successfully.

configure.ac
pablooliveira: Ok !

pablooliveira (approved):

pablooliveira:
@titeup PR approved. Thanks !

PR #339, opened 2023-03-19
New Interflop API

yohanchatelain:


titeup:
We need to start this review... 
@pablooliveira shall we meet to organize that? I think we were planning to do it by synchronous  chunk. I have more time starting next week.

pablooliveira:
> We need to start this review... @pablooliveira shall we meet to organize that? I think we were planning to do it by synchronous chunk. I have more time starting next week.

I will not have the available bandwidth to work on this until after mid May.
But by no means should you wait for me to start the review process.

titeup:
> > We need to start this review... @pablooliveira shall we meet to organize that? I think we were planning to do it by synchronous chunk. I have more time starting next week.
> 
> I will not have the available bandwidth to work on this until after mid May. But by no means should you wait for me to start the review process.

I will start by looking to the FMA addition and test it, this remove the flag to prevent FMA generation we introduce for latest llvm.

pablooliveira:
> I will start by looking to the FMA addition and test it, this remove the flag to prevent FMA generation we introduce for latest llvm.

Sounds good, thanks :pray:

install-interflop.sh
```diff
@@ -0,0 +1,72 @@
+#!/bin/bash
+
+REPO=interflop
```
pablooliveira, line 3: I'm not sure we should necessarily depend on the external interflop repository. Will it be maintained in the long run ?

If I understand correctly, for the moment this is only to retrieve the common header files ? The backends seem to be still be handled within verificarlo.

Moreover, on the technical side, why use here direct git commands instead of submodules (such as what we did with significantdigits) ?

pyproject.toml
```diff
@@ -0,0 +1,62 @@
+[build-system]
+requires = ["hatchling", "pathspec"]
+build-backend = "hatchling.build"
+
+[project]
+name = "verificarlo"
+version = "0.9.1"
```
pablooliveira, line 7: Can we retrieve the version number from configure.ac automatically, it simplifies releases. Now we have to change the version number only in two places: configure.ac and README.md. I would prefer we keep it like that.

src/backends/Makefile.am
```diff
@@ -1 +0,0 @@
-SUBDIRS= interflop-ieee interflop-mca interflop-mca-int interflop-cancellation interflop-bitmask interflop-vprec
```
pablooliveira, line 1: Why is this removed ?

src/interflop-stdlib/automake-expand
```diff
@@ -0,0 +1,118 @@
+#!/usr/bin/env python3
```
pablooliveira, line 1: When and how is automake-expand used ? I'm not sure to understand.

pablooliveira (commented):
Hi @yohanchatelain, as discussed I'm starting the review process on this PR. If I understood correctly, the idea is to make it land on `main` around January/February 2024.

This first pass is not extensive, I am mainly raising some questions regarding the integration with the `interflop` external repository.

install-interflop.sh
yohanchatelain: Yes, you're right. I included the interflop-stdlib in Verificarlo.

pyproject.toml
yohanchatelain: Yes, you're right. I updated it, so now the version is in the `VERSION` file. It's read by `configure.ac` and `pyproject.toml`. We still have to update it manually in the `README.md` since there is no mechanism to read a file from it.

src/backends/Makefile.am
yohanchatelain: In the previous setup, the backends were installed separately during the execution of the `install-interflop.sh` script. However,  I think it's more logical to handle the backend installation directly within verificarlo itself.

Now, each backend comes with its own configure script. These scripts are automatically executed as part of Verificarlo's main configure process. Specifically, the `ax_interflop_backends.m4` script located in `src/interflop-stdlib/m4/`  scans the `src/backends` directory to identify all available backends and then populates the `INTERFLOP_BACKENDS` variable with this information. The backends are dynamically recognized and included without the need to manually modify the SUBDIRS variable. This change simplifies the process of adding new backends, as they are now automatically detected and integrated during configuration.

src/interflop-stdlib/automake-expand
yohanchatelain: This script was used to populate the interflop-config script, but I ultimately decided to remove it since it would add an extra layer of complexity that is not necessary. The idea was to have something similar to llvm-config, but I think it's overkill.

pablooliveira:
Thanks @yohanchatelain ! This looks great, thanks for answering my points above and simplifying the installation process.

I think the code is very nice and it could be merged soon. I would like to discuss some issues live, would you be available for a call?

Thanks :pray:

src/backends/Makefile.am
pablooliveira: :+1:

src/backends/interflop-backend-vprec/common/vprec_tools.c
```diff
@@ -0,0 +1,325 @@
...
+ *     CMLA, Ecole Normale Superieure de Cachan                              *\
+ *                                                                           *\
+ *  Copyright (c) 2018                                                       *\
+ *     Universite de Versailles St-Quentin-en-Yvelines                       *\
+ *                                                                           *\
+ *  Copyright (c) 2019-2021                                                  *\
+ *     Verificarlo Contributors                                              *\
+ *                                                                           *\
+ ****************************************************************************/
+#include <math.h>
```
pablooliveira, line 20: For VPREC, for t > 27 we cannot guarantee correct rounding. See Samuel A. Figueroa. “When is Double Rounding Innocuous?” In: SIGNUM
Newsl. 30.3 (1995), pp. 21–26. issn: 0163-5778. doi: 10.1145/221332.
221334. url: https://doi.org/10.1145/221332.221334.

pablooliveira, line 20: For VPREC, for t > 27 we cannot guarantee correct rounding. See 
Samuel A. Figueroa. “When is Double Rounding Innocuous?” In: SIGNUM
Newsl. 30.3 (1995), pp. 21–26. issn: 0163-5778. doi: 10.1145/221332.
221334. url: https://doi.org/10.1145/221332.221334.

pablooliveira: and page 158 in my habilitation manuscript. Could you please git rebase and edit the message log?

doc/02-Backends.md
```diff
@@ -342,7 +351,7 @@ developpement.
 ### VPREC Backend (libinterflop_vprec.so)
 
 The VPREC backend simulates any floating-point formats that can fit into
-the IEEE-754 double precision format with a round to the nearest.
+the IEEE-754 double precision format with a round to the nearest ties to even.
```
pablooliveira, line 354: Check previous remark about faithful rounding for t > 26.

src/interflop-stdlib/fma/interflop_fma.c
```diff
@@ -0,0 +1,17 @@
+#include <quadmath.h>
```
pablooliveira, line 1: Can we please use the shim header at `test_bitmap_backend/pfp128.h` ? Libquadmath is not available on aarch64. So we should use the fma implementation in the shim and avoid linking with lquadmath for arm64.

pablooliveira (approved):
Dear @yohanchatelain as discussed together, I think we can merge this PR.

There are a few points remaining, we took notes during our call.

As you said, we should make casts instrumentation optional and disabled by default to avoid surprises for historical verificarlo users.

pablooliveira:
Looks great !

PR #340, opened 2023-04-13
WIP: Porting verificarlo to aarch64 (arm64)

### LINKED ISSUE #332: Support for aarch64 / arm64
Hello, I habe a problem when I follow the installation instruction on aarch64 sturcture. When I execute ./configure --with-flang=no, gcc-7,g++-7  for the latest version. It will check the __float128. And the result shows that __float128 is not supported. So, how can I solve this problem? 
I am using aarch64 and Ubuntu18.04 pc 
Thanks!

pablooliveira:
This PR tracks efforts to port verificarlo to aarch64 (arm64) / Linux platforms. This should fix #332.

I am working on the port thanks to an access to Turpan (https://www.calmip.univ-toulouse.fr/espace-utilisateurs/doc-technique-turpan) kindly provided by CALMIP. Nodes are equipped with Ampere Altra Q80-30 processors and RHEL is installed.

A first issue is that `__float128` type is not supported on aarch64. Fortunately, it appears that _Float128 (specified in the C programming language, ISO/IEC TS 18661) is supported by GCC both in amd64/linux and aarch64/linux.
A first patch, included here, swaps the types. Thanks to this patch I can build and install verificarlo on aarch64 and run basic tests.

We might also have issues with the call stack code used in the funcInstrumentation pass. I will tackle this next.


pablooliveira:
All tests pass except `test_absolute` which uses `src/common/float_struct.h` (with `_Float128`). Unfortunately, `clang` does not support `_Float128` . I think the easiest way forward is rewrite the test so that the test harness is compiled with GCC instead.

PR #348, opened 2024-10-10
Document mca_int vs mca differences in the user's documentation

pablooliveira:
After the discussion (https://github.com/verificarlo/verificarlo/discussions/347) it becomes clear that it is important to better document the differences and compromises between the different MCA backends in the user's doc.


yohanchatelain (approved):
LGTM! Thanks

PR #349, opened 2024-11-26
Separate make and make install

pablooliveira:
Currently `make` triggers the installation of the `interflop-stdlib` libraries.
Therefore when running `make` on a system-wide target directory the install commands fail for lack of permissions.

This PR separates the build and install phases for interflop-stdlib. Thus one can do
```
make
sudo make install
```
building as a normal user and installing as root.

@yohanchatelain could you please check that this does not break the new installation process? It does seem to work fine on my side.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -1,10 +1,6 @@
 ACLOCAL_AMFLAGS=-I m4
 
-all: interflop-stdlib-install $(SUBDIRS)
-
-interflop-stdlib-install:
-	make -C src/interflop-stdlib
-	make -C src/interflop-stdlib install
+all: $(SUBDIRS)
 
 SUBDIRS=src/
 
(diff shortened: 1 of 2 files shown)


pablooliveira:
Oh. It seems we need to install first the interflop headers. That is unfortunate... It worked on my system, since they were already installed from a previous version.

I'm not sure how to separate build and install process cleanly given this. Maybe we could have separate targets and document this in the install manual?

```
sudo make install-interflop-stdlib
make && sudo make install 
```

yohanchatelain:
Hi @pablooliveira, I originally hid the installation of `interflop-stdlib` to simplify things for users, but it seems to be causing more problems in the long run. I think it would actually be better to make the installation process visible.

Additionally, many of our users now use Verificarlo via Docker, so we (the devs) are practically the only ones installing it manually anymore lol.

yohanchatelain (approved):
LGTM! Thanks for debugging that.

PR #350, opened 2025-01-20
Add PRISM backend into Verificarlo

yohanchatelain:
This backend provides probabilistic rounding for stochastic rounding (sr) and up-down modes. It provides vectorized implementation for each mode as well as a dynamic and static dispatch. Dynamic dispatch selects the best implementation at runtime while static has fixed ISA. Instrumentation is done directly, meaning the backend cannot be changed with VFC_BACKENDS.

- Add specific instrumentation pass for this backend (src/libvfcinstrumentprism)
- Add prism backend rounding modes options (--prism-backend {up-down,stochastic-rounding})
- Add prism backend dispatch option (--prism-backend-dispatch {static,dynamic})
- Add prism backend debug options (--prism-backend-debug [{getoperands,abi,targetfeatures}])
- Add test_prism_backend to test both modes with dynamic/static dispatch for scalar and vector types
- Update documentation for PRISM backend mode
- Add bytecode generation (.bc) by default for verificarlo instrumentation pass to save compilation time. .ll files still generated when args.save_temps used.
- Update Dockerfile to install bazelisk (used to build PRISM)


yohanchatelain:
Thanks @pablooliveira!

I added a reference to PRISM SR in the mcaint backend documentation.

pablooliveira:
Thanks for adding the cross reference. Yes, I think a new release is indeed a good idea!

PR #351, opened 2025-01-21
Fix links in documentation

pablooliveira:
Just saw that the link to the PRISM repository in the documentation is broken.


yohanchatelain (approved):
Good catch!

PR #375, opened 2026-06-09
Copy VFC_BACKENDS before parsing

pablooliveira:
strtok_r modifies in place vfc_backends, adding \0 after each word, breaking arguments for child processes.

To avoid this issue, we copy the string to a new buffer before parsing.


yohanchatelain (approved):
Nice catch!

PR #378, opened 2026-06-30
interflop-backend-prism: support prism set_rounding_mode

pablooliveira:
Add support for prism set_rounding_mode() both as a user-call and through backend args.


yohanchatelain (approved):
LGTM!

PR #380, opened 2026-07-22
ddebug: disable debuginfod in inclusion generation

### LINKED ISSUE #379: Deadlock in test_ddebug_archimedes due to ubuntu debuginfo.
Found it with Claude.

`test_ddebug_archimedes` isn't nondeterministically flaky — it deadlocks every time during the reference run.

Root cause chain:

1. `vfc_ddebug`'s reference run executes `./archimedes` (built with `--ddebug`) via `ddRun`.
2. At program exit, verificarlo's `vfc_atexit()` destructor calls `ddebug_generate_inclusion()` (`src/vfcwrapper/vfcwrapper.c:169-195`, generated from `src/vfcwrapper/main.c.in`). For every instrumented FP op it needs to log, it `fork()`s and `execlp()`s:
`execlp(ADDR2LINE_BIN, ADDR2LINE_PATH, "-fpaCs", "-e", executable, addr, NULL);`
2. then the parent just calls `wait(&status)` — no timeout.
3. Your environment has `DEBUGINFOD_URLS=https://debuginfod.ubuntu.com` set. LLVM 18's `llvm-addr2line` honors that automatically and, even though it's given an address directly on the command line, it also fires off a debuginfod HTTP lookup for extra debug info.
4. In this sandbox, that outbound connection is silently dropped rather than refused, so `llvm-addr2line` blocks forever in `poll()` waiting on the socket. I confirmed this directly:
`llvm-addr2line -fpaCs -e ./archimedes 0x402454`        → hangs
`llvm-addr2line -fpaCs --no-debuginfod -e ./archimedes 0x402454` → "archimedes at archimedes.c:17" instantly
5. Since the parent (`archimedes`) just `wait()`s on that child, the whole reference-run subprocess hangs forever, `vfc_ddebug` never gets a return code, and the `Makefile:9: dd` target sits until someone Ctrl-C's it — exactly the `KeyboardInterrupt` in `test.log`.

**Immediate unblock** (no source change): unset `DEBUGINFOD_URLS` before running the test — I verified this makes `llvm-addr2line` return immediately.

**Real fix**: add `--no-debuginfod` to the `execlp` call in `src/vfcwrapper/main.c.in` (and regenerate `main.c/vfcwrapper.c`), since verificarlo always points addr2line at a local `/proc/<pid>/exe` and never wants a network round-trip here.

yohanchatelain:
Pass --no-debuginfod to addr2line in ddebug inclusion generation to avoid hangs when DEBUGINFOD_URLS points to an unreachable endpoint. Add a timeout-based regression check in test_ddebug_gen_include that sets DEBUGINFOD_URLS to an invalid URL and fails quickly if generation blocks.

Fix #379

### DIFF
--- a/src/vfcwrapper/main.c.in
+++ b/src/vfcwrapper/main.c.in
@@ -179,8 +179,8 @@ void ddebug_generate_inclusion(char *dd_generate_path, vfc_hashmap_t map) {
                  (void *)(get_value_at(map->items, i) - CALL_OP_SIZE));
         snprintf(executable, 64, "/proc/%d/exe", getppid());
         dup2(output, 1);
-        execlp(ADDR2LINE_BIN, ADDR2LINE_PATH, "-fpaCs", "-e", executable, addr,
-               NULL);
+        execlp(ADDR2LINE_BIN, ADDR2LINE_PATH, "-fpaCs", "--no-debuginfod", "-e",
+               executable, addr, NULL);
         logger_error("error running " ADDR2LINE_BIN);
       } else {
         int status;
--- a/tests/test_ddebug_gen_include/test.sh
+++ b/tests/test_ddebug_gen_include/test.sh
@@ -2,7 +2,27 @@
 set -e
 verificarlo-c --ddebug -g -O0 test.c -o test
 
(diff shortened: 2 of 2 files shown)


pablooliveira (approved):
LGTM ! Thanks

